PR #5, opened 2022-05-05
add SubPrefix method to ConfigArrayPrefix

awrichar:


pkg/config/config.go
```diff
@@ -330,6 +350,26 @@ func (c *configPrefix) AddKnownKey(k string, defValue ...interface{}) {
...
+		c.parent.AddChild(key, defValue...)
+	}
+}
+
+func (c *configPrefixArray) AddChild(k string, defValue ...interface{}) {
+	// When a child is added anywhere below this array, add it to the defaults map
+	prefix := c.base + "[]."
+	c.defaults[strings.TrimPrefix(k, prefix)] = defValue
+
+	// Note that arrays of arrays are not supported, so we don't bubble upwards
```
peterbroadhurst, line 364: What about:

`ffresty.InitPrefix(arrayPrefix.SubPrefix("blockchain").SubPrefix("ethconnect"))`
... would `AddKnownKey` on the `ethconnect` double-sub'd prefix bubble up ok?

peterbroadhurst: aha - answered my own question. Yes. `AddChild` on `configPrefix` does bubble up 👍

peterbroadhurst (approved):
Overall I feel like the terminology we (well I 😳 ) used for `prefix`/`subPrefix` hasn't proved to be super self-explanatory over time. The journey it took to get to this (really nice and compact) fix for the problem would probably have been shorter if the terminology was updated.

However, that's not in any way a blocker to merging this, and I don't have any great proposals for different terminology.

awrichar:
Yea, 100% agreed on the terminology. Will continue to ponder it as I know we'll likely be making more config enhancements in the coming weeks.

PR #6, opened 2022-05-06
Rename config.Prefix to config.Section

awrichar:


pkg/wsclient/wsconfig.go
```diff
@@ -53,7 +53,7 @@ func InitPrefix(prefix config.KeySet) {
 	prefix.AddKnownKey(WSConfigHeartbeatInterval, defaultHeartbeatInterval)
 }
 
-func GenerateConfigFromPrefix(prefix config.Prefix) *WSConfig {
+func GenerateConfigFromPrefix(prefix config.Section) *WSConfig {
```
peterbroadhurst, line 56: I think all the `prefix` variable names should change too, here and as we roll this out.

peterbroadhurst (changes requested):
Really like the new terminology... sorry to add to the work list with my request

peterbroadhurst (approved):
👍

PR #7, opened 2022-05-16
Add language tag to FFM, FFE, FFC, etc

nguyer:
This PR adds a language tag to all of our strings. This allows us to register the same message key for multiple languages. It also updates the `Expand` and `ExpandWithCode` functions to use the current language specified in the context, or fallback to an en_us string.


pkg/i18n/en_base_config_descriptions.go
```diff
@@ -29,70 +31,70 @@ var IgnoredType = "IGNORE"
 
 //revive:disable
 var (
-	ConfigGlobalConnectionTimeout = FFC("config.global.connectionTimeout", "The maximum amount of time that a connection is allowed to remain with no data transmitted", TimeDurationType)
-	ConfigGlobalRequestTimeout    = FFC("config.global.requestTimeout", "The maximum amount of time that a request is allowed to remain open", TimeDurationType)
+	ConfigGlobalConnectionTimeout = FFC(language.AmericanEnglish, "config.global.connectionTimeout", "The maximum amount of time that a connection is allowed to remain with no data transmitted", TimeDurationType)
```
awrichar, line 34: Should we create an alias function at the top of this file, rather than writing the language on every line? Feels ripe for a copy-paste error in the future.

awrichar, line 34: Should we create an alias function at the top of this file, rather than writing the language on every line? Feels ripe for a copy-paste error in the future.

awrichar: Might also consider a separate sub-package per language?

I'm actually not clear on how the other language vars will be added and used. Do the var names all need to be unique for each language?

nguyer: I like the suggestion for creating an alias at the top. I'll make that change. I'll try out sub packages, but I'm not sure how well that will work. If they're in the same package, all the language var names need to be unique, yes.

awrichar (approved):

PR #8, opened 2022-05-17
Add support for config with arrays of arrays

awrichar:


peterbroadhurst (approved):
Thanks @awrichar - think I worked out how this solves the problem of nested arrays, by flattening up to the last array in heirarchy.

PR #15, opened 2022-06-01
Remove FFCAPI for updated architecture

peterbroadhurst:
When events are involved, the coupling between the FFTM and the connector code is just too intertwined to support separate runtime processes. So a refactor of the FFCAPI needs to happen, and it makes sense for that to happen within a refactored FFTM.

So it should be removed from here.


awrichar (approved):
👋

PR #19, opened 2022-06-17
Move FFI Code from Core

nguyer:
This PR moves `FFI` and related types to `firefly-common` out of the core repo, so that they can be re-used in other FireFly subprojects. Any FFI code that dealt with Ethereum specific requirements, including conversion to/from ABIs, has been moved to `firefly-signer`.


pkg/fftypes/fftypes.go
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package fftypes
+
+import (
+	"crypto/sha256"
+)
+
+func TypeNamespaceNameTopicHash(objType string, ns string, name string) string {
```
peterbroadhurst, line 23: Wonder about the filename for this file - would `hashutils.go` be clearer?

peterbroadhurst (approved):
Approval added with one comment to review before merge

pkg/fftypes/fftypes.go
nguyer: Sure that makes sense. Yeah I wasn't really sure where to put this one at first.

PR #21, opened 2022-06-24
Allow number and boolean as JSON schema options

peterbroadhurst:
See https://github.com/hyperledger/firefly-signer/pull/11

- Allows `number` for non-integer types (`fixed`/`ufixed`) that can be specified as non-string
- Update `boolean` support to include `true`/`false` strings as an option (all elementary types allow string basically)


nguyer:
This is necessary, but the changes in this PR allow several combinations that should not be allowed. For example, this allows a schema that accepts both a `boolean` and a `number`.  I've implemented it slightly differently in:

https://github.com/hyperledger/firefly-common/pull/22

PR #24, opened 2022-07-08
Auth plugin framework

nguyer:
This PR adds the auth plugin framework and a simple basic auth plugin to the common library which can be used by each FireFly microservice. In the common library, it is applied at the HTTP listener level. However, other services can use the Authorize function however they need. For example, FireFly Core will also use the same function at a namespace level on HTTP requests, and WebSocket messages.

The basic auth plugin expects a `passwordfile` config key to be set, which should point to a file created with htpasswd using bcrypt hashed passwords. For example, you can create such a file by running:
```
htpasswd -cB allowed_users firefly
```


pkg/auth/authfactory/factory.go
```diff
@@ -0,0 +1,67 @@
...
+
+const (
+	// ConfigName the name of the specific instance of a particular auth plugin
+	ConfigName = "name"
+	// ConfigType the auth plugin to use for the HTTP server
+	ConfigType = "type"
+)
+
+var pluginsByType = map[string]func() auth.Plugin{
+	(*basic.Auth)(nil).Name(): func() auth.Plugin { return &basic.Auth{} },
```
awrichar, line 36: This is a pattern propagated from core, but seeing it again now... is it worth just moving Name to package scope, ie `basic.Name()`? Maybe we still need the one on the plugin itself though...

Not critical, just wanted to raise it again as the casting does look odd.

pkg/auth/authfactory/factory.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func RegisterPlugins(plugins map[string]func() auth.Plugin, conf config.Section) {
+	for k, plugin := range plugins {
+		plugin().InitConfig(conf)
+		pluginsByType[k] = plugin
```
awrichar, line 42: Seems weird to initialize the map inline above and then overwrite it all here

pkg/auth/authfactory/factory.go
```diff
@@ -0,0 +1,67 @@
...
+
+func InitConfigArray(config config.ArraySection) {
+	config.AddKnownKey(ConfigName)
+	config.AddKnownKey(ConfigType)
+	for name, plugin := range pluginsByType {
+		plugin().InitConfig(config.SubSection(name))
+	}
+}
+
+func InitConfig(config config.Section) {
```
awrichar, line 54: I'm a bit confused by the inclusion of `RegisterPlugins`, `InitConfigArray`, _and_ `InitConfig`, all of which seem to do some variation of `plugin().InitConfig(...)`. Are these just trying to support multiple ways of using and initializing the plugin?

pkg/auth/basic/config.go
```diff
@@ -0,0 +1,32 @@
...
+package basic
+
+import (
+	"github.com/hyperledger/firefly-common/pkg/config"
+)
+
+const (
+	// BasicAuthConfigKey description
+	BasicAuthConfigKey = "basic"
+	// PasswordFile description
```
awrichar, line 26: Might want to fill in these placeholder comments

pkg/fftypes/auth.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"net/http"
+	"net/url"
+)
+
+type AuthReq struct {
+	Method    string
+	URL       *url.URL
+	Header    http.Header
+	Namespace string
```
awrichar, line 27: I see `Namespace` included here, but is it actually parsed by this library?

pkg/auth/basic/basic_auth.go
```diff
@@ -0,0 +1,87 @@
...
+
+type Auth struct {
+	users map[string][]byte
+}
+
+func (a *Auth) Name() string {
+	return "basic"
+}
+
+func (a *Auth) Init(ctx context.Context, name string, config config.Section) error {
```
awrichar, line 41: Is this `name` used?

pkg/httpserver/server_auth.go
```diff
@@ -0,0 +1,42 @@
...
+	if pluginName != "" {
+		authPlugin, err := authfactory.GetPlugin(ctx, pluginName)
+		if err != nil {
+			return nil, err
+		}
+		if err := authPlugin.Init(ctx, "", conf.SubSection(authPlugin.Name())); err != nil {
+			return nil, err
+		}
+		handler := &auth.Handler{}
+		handler.Init(authPlugin)
```
awrichar, line 38: Maybe consider a `NewHandler()` factory instead of this two-step init?

pkg/httpserver/server_auth_test.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/hyperledger/firefly-common/pkg/config"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestServerAuthBasicAuth(t *testing.T) {
+	config.RootConfigReset()
+	section := config.RootSection("http")
+	InitHTTPConfig(section, 5000)
+	authConfig := section.SubSection("auth")
+	// authfactory.InitConfig(authConfig)
```
awrichar, line 33: Can the commented lines be deleted?

pkg/auth/basic/basic_auth.go
```diff
@@ -0,0 +1,87 @@
...
+	users, err := readPasswordFile(passwordFile)
+	if err != nil {
+		return err
+	}
+	a.users = users
+	return nil
+}
+
+func (a *Auth) Authorize(ctx context.Context, req *fftypes.AuthReq) error {
+	authHeader := strings.TrimPrefix(req.Header.Get("Authorization"), "Basic ")
```
awrichar, line 52: Do you want to check if the prefix was actually stripped? If they pass something other than `Basic` auth, are you just relying on the later check (line 55) to catch that?

pkg/auth/authfactory/factory.go
nguyer: This function shouldn't be overwriting it. This is for other microservices (e.g. core) to have a way to register plugins implementations outside of the common library and still have this factory be able to instantiate them.

pkg/auth/authfactory/factory.go
nguyer: - RegisterPlugins is for code that includes this library to be able to add its own implementations of an Auth plugin.
- InitConfig creates known keys in a config structure that comes from an HTTP server config section and not an array
- InitConfigArray creates known keys in a config structure that comes from the list of a plugins (in FireFly core)

pkg/auth/basic/basic_auth.go
nguyer: It is not used by this plugin, but other plugin interfaces include it in their signature, so I included it here, in case a plugin down the road needs it.

pkg/auth/authfactory/factory.go
awrichar: Ah, makes sense. Think you could stick that in a comment here?

pkg/auth/authfactory/factory.go
awrichar: Maybe all of these should be doc comments in this file :)

awrichar (approved):
Requested that a few of your answers here be added to the code comments for posterity, but other than that this looks great.

PR #27, opened 2022-08-01
Utility for serializing structs into a JSON Map (or fftypes.JSONObject)

peterbroadhurst:
This utility lets you use the standard JSON serialization policy for fields in go structs, to write fields into a map.

This has two uses in FireFly to avoid horrible JSON marshal/unmarshal loops:

1. Combining multiple Go structures from plugins into an extensible JSON body - required in FFTM here https://github.com/hyperledger/firefly-transaction-manager/pull/14
2. Converting from Go structs to an `fftypes.JSONObject`

### DIFF
--- a/pkg/jsonmap/jsonmap.go
+++ b/pkg/jsonmap/jsonmap.go
@@ -1,0 +1,74 @@
+// Copyright © 2022 Kaleido, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


pkg/jsonmap/jsonmap.go
```diff
@@ -0,0 +1,80 @@
...
+
+func StructToJSONMap(i interface{}) map[string]interface{} {
+	m := make(map[string]interface{})
+	AddJSONFieldsToMap(reflect.ValueOf(i), m)
+	return m
+}
+
+// AddJSONFieldsToMap is a helper for marshalling struct fields down into a map
+//
+// Note: Does not currently respect the `omitempty` JSON flag semantics
```
nguyer, line 32: I think it actually does now 🙂

nguyer (approved):

PR #34, opened 2022-08-24
adding IsSet function for default value checking

Chengxuan:
For configuration migration. Exposing IsSet function so that we can check whether a key is manually set by the users.

Signed-off-by: Chengxuan Xing <chengxuan.xing@kaleido.io>


peterbroadhurst (approved):
👍

PR #35, opened 2022-09-02
Add FFEnumParseString for validation

nguyer:
Signed-off-by: Nicko Guyer <nicko.guyer@kaleido.io>

### DIFF
--- a/pkg/fftypes/enum.go
+++ b/pkg/fftypes/enum.go
@@ -54,3 +54,16 @@ func (ts *FFEnum) UnmarshalText(b []byte) error {
 	*ts = FFEnum(strings.ToLower(string(b)))
 	return nil
 }
+
+func FFEnumParseString(t string, i string) FFEnum {
+	e, ok := enumValues[t]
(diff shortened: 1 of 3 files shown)


pkg/fftypes/enum.go
```diff
@@ -54,3 +54,16 @@ func (ts *FFEnum) UnmarshalText(b []byte) error {
 	*ts = FFEnum(strings.ToLower(string(b)))
 	return nil
 }
+
+func FFEnumParseString(t string, i string) FFEnum {
+	e, ok := enumValues[t]
```
peterbroadhurst, line 59:
```suggestion
	e, ok := enumValues[strings.ToLower(t)]
```

Our standard semantic is case insensitivity. I think we

peterbroadhurst: Also, should we have a standard error for the case that it is non-empty, but not one of the valid values?

peterbroadhurst (approved):

PR #38, opened 2022-11-07
Go 1.18

peterbroadhurst:
Updates for go 1.18, including Mockery v2 which is go 1.18 compatible, and pulling in a range of freshened dependencies


mocks/authmocks/plugin.go
```diff
@@ -1,65 +0,0 @@
-// Code generated by mockery v1.0.0. DO NOT EDIT.
-
-package authmocks
```
nguyer, line 3: `authmocks` is currently used by unit tests in FireFly core in two places:
https://github.com/hyperledger/firefly/blob/4cdaad1d9b440451b97bf6c8a68b3d6f77fc530a/internal/orchestrator/orchestrator_test.go#L491
and
https://github.com/hyperledger/firefly/blob/b31f9a678e2f2c3db93e662c595d7a5eaabf4be4/internal/namespace/manager_test.go#L67

Do we need to add this back, or were those code sections also changed in core in a corollary PR?

peterbroadhurst: Sorry about that - not sure how I managed that. Re-ran `make mocks` to re-instate it.

pkg/ffapi/openapi3.go
```diff
@@ -282,7 +282,7 @@ func (sg *SwaggerGen) AddParam(ctx context.Context, op *openapi3.Operation, in,
 	}
 	var defValue interface{}
 	if def != "" {
-		defValue = &def
+		defValue = def
```
peterbroadhurst, line 285: Note this is required due to a tightening of requirements in the OpenAPI V3 package we use (`*string` not supported)

nguyer (approved):
Looks good! Thanks for updating this!

PR #41, opened 2022-12-19
Add support for passing custom HTTP headers through

nguyer:
This PR adds support for two new features that can be used in FireFly core and other microservices.

- A header named `X-FireFlyRequestID` can be set on HTTP requests and the service will use that as the internal request ID, rather than a generated short ID. The request ID is now also propagated to other FFResty client requests so it can be used as a trace ID through several services.
- Support for passing custom HTTP request headers through. A set of header names can be provided in the `ffapi.HandlerFactory` called `PassthroughHeaders`. In FireFly core, these are specified under the `api` config section. These headers will be put onto the request context.
- If an FFResty client is configured with `passthroughHeadersEnabled: true` in its config section, it will set any headers that have been packed into the context on the new request that it builds


awrichar (approved):

pkg/ffapi/handler.go
```diff
@@ -34,12 +34,20 @@ import (
 	"github.com/hyperledger/firefly-common/pkg/log"
 )
 
+const FFRequestIDHeader = "X-FireFlyRequestID"
```
awrichar, line 37: Could consider hyphenating this - ie `X-FireFly-Request-ID`

awrichar: (I think that's the general convention for multi-word HTTP headers)

PR #48, opened 2023-02-22
Add ability to register custom message catalogs

peterbroadhurst:
Currently we have a source-code controlled list of message catalogs for core FireFly components, to avoid clashes.

However, this precludes use of the utility functions outside of core FireFly codebases - such as in custom plugins under development. It's a useful library, and this restriction doesn't seem to make sense. Even worse we might end up in a position where code uses the wrong message catalog to define messages for convenience.

### DIFF
--- a/pkg/i18n/messages.go
+++ b/pkg/i18n/messages.go
@@ -1,4 +1,4 @@
-// Copyright © 2022 Kaleido, Inc.
+// Copyright © 2023 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
@@ -19,6 +19,7 @@ package i18n
 import (
 	"context"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"golang.org/x/text/language"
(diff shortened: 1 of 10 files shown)


pkg/i18n/messages.go
```diff
@@ -69,6 +70,28 @@ var msgIDUniq = map[string]bool{}
...
+// then you should instead submit a PR to update the registeredPrefixes constant.
+//
+// If you're writing a proprietary extension, or just using the Microservice framework for your
+// own purposes, then you can register your own prefix with this function dynamically.
+//
+// It must conform to being two upper case characters, then two numbers.
+// You cannot use the `FF` prefix characters, as they are reserved for FireFly components
+func RegisterPrefix(prefix, description string) {
+	if !prefixValidator.MatchString(prefix) || strings.HasPrefix(prefix, "FF") {
+		panic("invalid prefix")
```
nguyer, line 87: It might be useful to print the prefix that was duplicated in the panic message here and below

nguyer (approved):
One minor suggestion but looks good to me!

PR #49, opened 2023-02-28
Add getter for DB provider features struct field

matthew1001:
Looking into issue https://github.com/hyperledger/firefly/issues/1196 (now spun out into https://github.com/hyperledger/firefly/issues/1199) it appears that the `features` struct field is no longer visible to users of the package. Adding a getter in a similar way to the existing `ConnLimit()` getter makes the field accessible to read.


peterbroadhurst (approved):
👍

PR #50, opened 2023-03-08
Add configuration option to allow TLS cert verification to be disabled

matthew1001:


EnriqueL8:
Whilst updating the firefly-perf-cli with the mTLS changes I noticed that it using a commit pointing to this branch, so going to review this and see if it conflicts. Spoke to @matthew1001 that he will make the changes to the cli to pick up the latest version

pkg/wsclient/wsclient.go
```diff
@@ -96,6 +98,8 @@ func New(ctx context.Context, config *WSConfig, beforeConnect WSPreConnectHandle
 		wsdialer: &websocket.Dialer{
 			ReadBufferSize:  config.ReadBufferSize,
 			WriteBufferSize: config.WriteBufferSize,
+			//nolint:gosec
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: config.DisableTLSVerification},
```
EnriqueL8, line 102: Unfortunately this doesn't play nicely with a secure websocket server which is the case when mTLS is enabled.

What we actually want is the ability to provide a `tls` section under the WSClient or take a configuration from elsewhere

EnriqueL8 (commented):
I think we should get this in and then raise a separate PR to properly expose a TLS config on this wsclient to secure the HTTP Upgrade.

peterbroadhurst:
Closing this as I believe WS TLS was implemented in #67 
@matthew1001 - please feel to raise a new PR to add a `HTTPConfTLSInsecureSkipVerify` boolean config variable to `fftls` with suitable warnings in the doc-entry for the config var, if there's still a requirement for this (I could see dev testing of WebHook clients from FireFly being an example of a limited use case for disabling TLS cert validation).

PR #51, opened 2023-03-15
move file watcher to a common package

shorsher:
Moves the file watcher logic to a new package, `fswatcher`, for potential use outside of viper config watching


peterbroadhurst (approved):
This looks great - will allow re-use of this code in other microservices much more easily.

PR #53, opened 2023-03-20
Common metrics code

Chengxuan:
Added a metrics manager for defining and emitting metrics with `ff` prefixes and subsystem names. a subsystem can only use one of the two options for creating metrics:
1. create a metrics manager to define custom metrics manually (with an options to use default system labels)
2. use out-of-box predefined metrics for a common resource type (e.g. HTTP request)


peterbroadhurst (approved):
This looks great 👍

PR #57, opened 2023-03-25
Support text/plain content type

jimthematrix:
Useful for APIs that relies on envelope based authentication, for instance in many decentralized identity frameworks, which may become part of FireFly in the future.


jimthematrix:
using a different PR to squash the commits for proper signoff

PR #60, opened 2023-03-27
CRUD utility for building microservices on the DB layer, and BigInt query fix

peterbroadhurst:
There's a lot of boilerplate still needed when building a microservice on top of `firefly-common` DB utilities.

This hasn't been addressed in FireFly Core, because so few of the collection are completely "standard", they need all kinds of special tweaks on a per collection basis. So the code is justified.

But in microservices like EVMConnect trying to pick up the framework, it would be great to just define the metadata of the collection and get the basic set of functions for free.

... so here's what's needed after this change (for SQL based DBs at least) - much better I hope:
https://github.com/hyperledger/firefly-common/blob/0d471d9d062ec110bd6426472a524e4ab2b8d440/pkg/dbsql/crud_test.go#L62-L96

I also found there was no `BigIntField` definition in the query support of `ffapi` - which seemed like a bad omission, particularly las we store the DB data in hex. A follow-on change would be needed in FF Core to pick that up for all `FFBigInt` fields

> In draft as I have a little remaining test coverage to close out on


nguyer (approved):

peterbroadhurst:
Extended the interface to support `JOIN` (on read - not write) through through some low-level hooks, and ability to modify the table alias:
https://github.com/hyperledger/firefly-common/blob/c35a11c2ad508b548e04462050f2bf637366e710/pkg/dbsql/crud_test.go#L126-L172

PR #61, opened 2023-03-29
Add `APIServer` utility, `showconf` Cobra command, and debug server

peterbroadhurst:
More items that can become common and avoid boilerplate in microservices built on this framework:
- A fully fledged `APIServer` that gives an opinionated way to run an API Server
   - Builds on the `httpserver` and `ffapi.Route`/`ffapi.Handler` framework making some opinionated decisions
   - Uses generics to allow you to pass through a single `manager` instance, or generate a context struct for each request
   - Includes metrics using the new metrics utilities recently added to common
- Cobra command for `showconfig` that you can just add to your root command
- Function to run the debug server with consistent config

Example of the code needed now to create an API Server:
https://github.com/hyperledger/firefly-common/blob/e6af84602ced0f11c35c2e8efbbd2ef287e2d702/pkg/ffapi/apiserver_test.go#L99-L113


pkg/ffapi/favicons.go
```diff
@@ -0,0 +1,45 @@
...
+)
```
nguyer, line 29: Another option would be to use the embed directive here and put the image files in the same package, rather than actually putting the hex here in the source

nguyer (approved):

nguyer:
Looks good! A question and a suggestion, but I'm happy for this to be merged as-is if you want to.

pkg/config/cobracmd.go
```diff
@@ -0,0 +1,43 @@
...
+
+package config
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+)
+
+func ShowConfigCobraCommand(initConf func() error) *cobra.Command {
```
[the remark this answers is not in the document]
peterbroadhurst: 👍 - I can change this

pkg/ffapi/favicons.go
peterbroadhurst: 👍 - I'll also make them configurable

peterbroadhurst: Done - I've extended the code I copied to:
- Have proper 16px and 32px FireFly PNG icons freshly built from SVG
- Use `// go:embed` notation
- Be customizable in the `APIServerOptions` structure

peterbroadhurst:
Applied all the suggestions - thanks @nguyer

PR #62, opened 2023-03-30
Expose the mock provider for other packages to use in tests

peterbroadhurst:


.vscode/settings.json
```diff
@@ -25,6 +25,7 @@
     "Infof",
     "Kaleido",
     "microservices",
+    "Nilable",
```
nguyer, line 28: I don't know why but this really bothers me... I feel like it should have two `l`s.

nguyer (approved):

.vscode/settings.json
peterbroadhurst: Might be a touch of ni[h]ilism

peterbroadhurst: (... going to make the change)

PR #64, opened 2023-04-17
Add "networkName" and "published" to FFI fields

awrichar:
These will be needed as part of https://github.com/hyperledger/firefly/issues/1220, based on its current trajectory. See notes there for an overview of plans and unanswered questions.

The argument could be made that these fields may or may not belong in the base `fftypes.FFI` struct, and that some of this info is metadata specific to the functionality in FireFly core, and that perhaps the FFI type should be wrapped there. That would generate more work, but I wasn't sure how "pure" we want to be about adding extra fields to this struct.


peterbroadhurst (approved):
I think this is pragmatic - and I don't feel a need to exclude these fields from the FFI at this point.

PR #65, opened 2023-04-19
feat: Add mTLS configuration for ffresty

EnriqueL8:
Adding the ability to supply a config to the ffresty client to setup TLS and mTLS.

**NOTE** This changes the interface of that init function to now return an error as well, so every consumer of this library and that client on update will need to handle that error. We could avoid doing this by allowing the user to provide a TLS Config as an argument but the pattern seems to be to provide viper config!

For testing I added a series of real and fake certificates and keys. I also explore standing up an mTLS mock server with the `httptest` but it proved cumbersome. I think validating that the correct certificates are added to the TLS Config should be good enough but happy to rectify


test/certs/ca-crt.pem
peterbroadhurst, line 1: Would be good to be consistent with this code - in fact to re-use that code.

The problem with checking in certs, is that they expire over time. So generating them is much preferred

https://github.com/hyperledger/firefly-common/blob/f43b8dafc0dcd153ffaaac6c0b44b0ad1ead55ef/pkg/httpserver/httpserver_test.go#L139-L163

pkg/ffresty/ffresty.go
```diff
@@ -92,6 +93,54 @@ func New(ctx context.Context, staticConfig config.Section) *resty.Client {
 			TLSHandshakeTimeout:   staticConfig.GetDuration(HTTPTLSHandshakeTimeout),
 			ExpectContinueTimeout: staticConfig.GetDuration(HTTPExpectContinueTimeout),
 		}
+
+		// Check for TLS is enabled
+		if staticConfig.GetBool(HTTPTLSEnabled) {
+			tlsConfig := &tls.Config{MinVersion: tls.VersionTLS12}
```
peterbroadhurst, line 99: Related to the other comments on checking in test files, would be good to consider if we can avoid duplication of all this code (and maybe even the config structure) with https://github.com/hyperledger/firefly-common/blob/f43b8dafc0dcd153ffaaac6c0b44b0ad1ead55ef/pkg/httpserver/httpserver.go#L111-L126

pkg/ffresty/ffresty.go
```diff
@@ -67,10 +70,7 @@ func OnAfterResponse(c *resty.Client, resp *resty.Response) {
 //
 // You can use the normal Resty builder pattern, to set per-instance configuration
 // as required.
-func New(ctx context.Context, staticConfig config.Section) *resty.Client {
-
-	var client *resty.Client
-
+func New(ctx context.Context, staticConfig config.Section) (client *resty.Client, err error) {
```
peterbroadhurst, line 73: As you mention, this line will drive a lot of change in the components.

I wonder if the right answer here, is to prepare a set of PRs against the other `hyperledger/firefly-*` repositories - otherwise they would become unable to pick up security maintenance.

peterbroadhurst (changes requested):
This is great @EnriqueL8 - A few suggestions/questions to consider.

test/certs/ca-crt.pem
EnriqueL8: Makes sense - will rework that!

pkg/ffresty/ffresty.go
EnriqueL8: Agreed - maybe a tls sub directory with it's own config

pkg/ffresty/ffresty.go
EnriqueL8: Sounds like a plan, have already started on the main firefly but will look at the other repos and coordinate the changes

pkg/ffresty/config.go
```diff
@@ -88,4 +91,7 @@ func InitConfig(conf config.KeySet) {
 	conf.AddKnownKey(HTTPExpectContinueTimeout, defaultHTTPExpectContinueTimeout)
 	conf.AddKnownKey(HTTPPassthroughHeadersEnabled, defaultHTTPPassthroughHeadersEnabled)
 	conf.AddKnownKey(HTTPCustomClient)
+
+	tlsConfig := conf.SubSection("tls")
+	fftls.InitTLSConfig(tlsConfig)
```
EnriqueL8, line 96: I wonder if this is the best way to do it - having problem with existing plugins in firefly to refactor the InitConfig

EnriqueL8: Resolve it to be a deprecated config we can remove

peterbroadhurst (approved):
Awesome. Thanks for going the extra mile to package this one all up so tidily @EnriqueL8!

PR #69, opened 2023-05-18
TLS Client Auth - Subject DN Validation

peterbroadhurst:
In PR chain with #68 

Provides subject distinguished name validation option, so that in addition to specifying the set of CAs that are valid, a set of match criteria can be applied to the subject of the certificate.


pkg/fftls/fftls.go
```diff
@@ -93,8 +96,103 @@ func ConstructTLSConfig(ctx context.Context, conf config.Section, tlsType string
...
+		attr = strings.ToUpper(attr)
+		if _, knownAttr := SubjectDNKnownAttributes[attr]; !knownAttr {
+			return nil, i18n.NewError(ctx, i18n.MsgInvalidTLSDnMatcherAttr, attr)
+		}
+		validatorString, ok := v.(string)
+		if !ok {
+			return nil, i18n.NewError(ctx, i18n.MsgInvalidTLSDnMatcherType, attr, v)
+		}
+		// Ensure full string match with all regexp
+		validatorString = "^" + strings.TrimSuffix(strings.TrimPrefix(validatorString, "^"), "$") + "$"
```
EnriqueL8, line 162: I'm confused about this since you are removing the prefix `^` and removing the suffix `$` and then adding them back again. I get that you want to remove as you don't want to add two `^^` or `$$` but if the validatorString starts with `^^` you still end up with `^^`.  Maybe not that important as we control the input here

EnriqueL8 (commented):
Looks good - just one tiny comment

pkg/fftls/fftls.go
peterbroadhurst: Yeah - it was just to allow users to end up with an identical result whether they specify:
- `^myreg$`
- `myreg`

I wanted to avoid the accidental security exposure of someone assuming a full match, but actually only having a partial match. Oddly Go doesn't seem to have a function on regex to specify the match must match the whole input string.

pkg/ffapi/favicons_test.go
```diff
@@ -28,7 +28,7 @@ import (
 func TestFavIcon16(t *testing.T) {
 	res := httptest.NewRecorder()
 	req := httptest.NewRequest("GET", "/api/favicon-16x16.png", nil)
-	var handler http.HandlerFunc = favIconsHandler(ffLogo16, ffLogo32)
+	var handler http.HandlerFunc = favIconsHandler(nil, nil)
```
nguyer, line 31: Just want to make sure that setting this to `nil` was an intentional choice here

pkg/i18n/en_base_config_descriptions.go
```diff
@@ -64,12 +64,13 @@ var (
...
-	ConfigGlobalTLSHandshakeTimeout = ffc("config.global.tlsHandshakeTimeout", "The maximum amount of time to wait for a successful TLS handshake", TimeDurationType)
+	ConfigGlobalTLSCaFile               = ffc("config.global.tls.caFile", "The path to the CA file for TLS on this API", StringType)
+	ConfigGlobalTLSCertFile             = ffc("config.global.tls.certFile", "The path to the certificate file for TLS on this API", StringType)
+	ConfigGlobalTLSClientAuth           = ffc("config.global.tls.clientAuth", "Enables or disables client auth for TLS on this API", StringType)
+	ConfigGlobalTLSEnabled              = ffc("config.global.tls.enabled", "Enables or disables TLS on this API", BooleanType)
+	ConfigGlobalTLSKeyFile              = ffc("config.global.tls.keyFile", "The path to the private key file for TLS on this API", StringType)
+	ConfigGlobalTLSRequiredDNAttributes = ffc("config.global.tls.requiredDNAttributes", "A set of required subject DN attributes. Each entry is a regular expression, and the subject certificate must have a matching attribute of the specified type (CN, C, O, OU, ST, L, STREET, POSTALCODE, SERIALNUMBER are valid attributes)", MapStringStringType)
```
nguyer, line 72: Should we also have an example of what this regex should look like? Maybe in the FireFly docs?

nguyer (approved):
Looks good to me! Just a couple of minor questions/suggestions, but feel free to merge

pkg/i18n/en_base_config_descriptions.go
peterbroadhurst: It's a great suggestion. I don't think we have an TLS reference section in the docs, beyond these lines duplicated on each config section.

@EnriqueL8 - I wonder if this is something you'd like to do for the base to demonstrate an example config.
I'd be very happy to add in a piece on the use of `requiredDNAttributes`

EnriqueL8: Sure happy to pick this up

PR #70, opened 2023-05-31
feat: Add nested array config

EnriqueL8:
As part of another feature I want to be able to add nested arrays such as:
```
namespaces:
  predefined:
  - name: myns
    tlsConfigs:
    - name: myconfig
      tls:
        enabled: true
```

This is not currently possible and this PR adds that support. Let me know if I missed out on anything!

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -354,20 +354,11 @@ func (c *configSection) SubArray(name string) ArraySection {
 }
 
 func (c *configArray) SubArray(name string) ArraySection {
-	a := &configArray{
+	return &configArray{
 		base:     keyName(c.base+"[]", name),
 		parent:   c,
 		defaults: make(map[string][]interface{}),
 	}
-	// Get defaults from any enclosing array entry, and copy over any applicable to this subtree
-	// This is necessary to propagate known keys for arrays within arrays
-	prefix := a.base + "[]."
-	for key, val := range getArrayEntryDefaults(c) {
-		if strings.HasPrefix(key, prefix) {
-			a.defaults[strings.TrimPrefix(key, prefix)] = val
-		}
-	}
-	return a
 }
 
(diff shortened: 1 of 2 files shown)


pkg/config/config.go
```diff
@@ -351,6 +353,23 @@ func (c *configSection) SubArray(name string) ArraySection {
...
+func (c *configArray) SubArray(name string) ArraySection {
+	a := &configArray{
+		base:     keyName(c.base+"[]", name),
+		parent:   c,
+		defaults: make(map[string][]interface{}),
+	}
+	// Get defaults from any enclosing array entry, and copy over any applicable to this subtree
+	// This is necessary to propagate known keys for arrays within arrays
+	prefix := a.base + "[]."
+	for key, val := range getArrayEntryDefaults(c) {
```
awrichar, line 365: I pulled this locally to test out, and noticed this block was lacking code coverage. I think that's because it shouldn't be necessary.

There are two ways of traversing the config tree, which confusingly overlap. One is to _build_ the list of allowed config values, in which case you might nest a SubArray within a SubArray. In this mode, there will be no ArrayEntries, so this list will always be empty.

The second is to _inspect_ the parsed config, in which case you always go from SubArray to ArrayEntry to SubArray. Therefore you'll never come through this method (because an ArrayEntry is a configSection, you'll be going via the other method above).

So I'm fairly certain you can drop this loop altogether, since you'll never realistically have a case where you call configArray.SubArray and _also_ have an ArrayEntry parent in the tree.

awrichar (approved):
I pondered over this for quite a while and I think it's good. I noted a block of code which I believe is dead and can be removed.

pkg/config/config.go
EnriqueL8: Ah makes sense - great explanation, thanks!

PR #71, opened 2023-06-08
Allow any string for ID column, and support collections without an `updated` column

peterbroadhurst:
Couple of updates/enhancements found to be required while working on FFTM/EVMConnect:

1. Allowing any string for ID
  - The default example `ResourceBase` still uses fftypes.UUID
  - This allows FFTM to use the `namesace:UUID` convention from FF Core (established historically)
2. Allowing collections without an `updated` column
  - The previous code allowed you to map the `created`/`updated` columns to any fields in your business object, but required them. This was a pain for write-or-delete-only data, which is true for performance sensitive transaction history records in FFTM/EVMConnect
3. Ensuring large integers are sortable in the DB
   - Uses zero padding on the already hex storage of the data. This was required for `nonce` sorting, but will have benefits in other places in the FireFly ecosystem. For example in Core it will be possible to sort account balances etc.
4. Add `DeleteMany` operation to CRUD
   - This wasn't added before because it can't generate events, but I found the lack of it was a big pain. So I've added it with the caveat that it still doesn't generate change events.
5. Add `Count` operation to CRUD
   - Just a simple omission as not predicted to be required before
6. Add features to support externalizing the `seq` column in flexible ways
   - An interface to set it into the object on insert/query
   - The ability to query the raw sequence integer in an efficient way for all objects
7. Allows REST API Filter for multiple conditions on a single value combine with `AND` optionally, rather than `OR` as is the default
    - This is required for a time range, with a start and an end
    - I went for using the`[` modifier for `AND` modifier, and and `]` for `OR` allowing you to do:
       - `?created=[>12345&created=[<23456`  - `AND`
       - `?created=[]>12345&created=[]<23456`  - `OR` (explicit - because second modifier overrides first)
       - `?created=]>12345&created=]<23456`  - `OR` (explicit)
       - `?created=>12345&created=<23456`  - `OR` (implicit)


pkg/cache/cache.go
```diff
@@ -78,6 +92,18 @@ func (c *CCache) Get(key string) interface{} {
 	return nil
 }
 
+// GetNotExpired retrieves from the cache, without extending the expiry time, and
```
awrichar, line 95: Looks like the name changed to "GetUnexpired"

pkg/cache/cache.go
```diff
@@ -28,6 +28,19 @@ import (
 // Manager contains functions to manage cache instances.
 // It provides functions for creating new cache instances and list all of the names of existing cache instances
 // Each cache instance has unique name and its own cache size and TTL configuration.
+//
+// Two modes of operation are available
+//
+//	Default behavior (when the normal Get() methods are used to access the cache):
+//	    - Item is added to the cache
+//	    - After the expiry time, it is automatically purged from the cache to release memory
+//	       - ** THIS PART IS NOT IMPLEMENTED, AND A REAPER NEEDS TO BE ADDED **
```
awrichar, line 37: Probably want to confirm @Chengxuan's thoughts, since he's been working on the cache a lot. Overall I think I'm OK with the description laid out here, but I want to be sure we're not overcomplicating this interface and conflating different meanings of the term "TTL".

awrichar (approved):

pkg/cache/cache.go
Chengxuan: PR raised here: https://github.com/hyperledger/firefly-common/pull/84

peterbroadhurst: This was all resolved in #86 in the end 👍

pkg/cache/cache.go
peterbroadhurst: This was resolved in #86

PR #73, opened 2023-06-08
feat: Add configuration struct for ffresty

EnriqueL8:


EnriqueL8:
Tests failing due to timeout errors on the ws tests - will take a look

EnriqueL8:
Actually I don't want to remove that test but there is a race condition introduced by https://github.com/hyperledger/firefly-common/commit/0024c5d6288e5cfd7ccd7b305535f49fdadb020b#diff-55f4fd757c9ababdff05013f114f9e63b2cbdd3533736fc20690948d823b7897R91

peterbroadhurst (approved):
Looks great, and thanks for making it so no changes are needed in other microservices that use this package (unless they want to use the feature).

PR #81, opened 2023-06-24
adding support for groupby

Chengxuan:
Add groupBy support for the sql filter builder


peterbroadhurst (approved):
👍 - so this lets internal code use GroupBy, but doesn't expose it on the REST APIs. That's great.

Issue #78, opened 2023-06-22
Fixing TTL in cache util


Chengxuan:
TTL settings are not functioning correctly as we figured out in https://github.com/hyperledger/firefly-common/pull/77

It should be fixed according to the design documented in the PR above.

Chengxuan:
TTL docs fix needs to be done by the consumer of this lib.

PR #84, opened 2023-06-29
proposal for ttl fix

Chengxuan:
for @awrichar 's review comment here: https://github.com/hyperledger/firefly-common/pull/71#pullrequestreview-1503466769

Made a breaking change in the cache interface of all Get methods to make the consumer aware of a bug so that they can now make a decision on how to deal with their existing TTL configurations if exposed.

Resolves https://github.com/hyperledger/firefly-common/issues/78

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -28,19 +28,6 @@ import (
 // Manager contains functions to manage cache instances.
 // It provides functions for creating new cache instances and list all of the names of existing cache instances
 // Each cache instance has unique name and its own cache size and TTL configuration.
-//
-// Two modes of operation are available
-//
-//	Default behavior (when the normal Get() methods are used to access the cache):
-//	    - Item is added to the cache
-//	    - After the expiry time, it is automatically purged from the cache to release memory
-//	       - ** THIS PART IS NOT IMPLEMENTED, AND A REAPER NEEDS TO BE ADDED **
-//	    - While it's in memory, it is returned from the Get() methods
-//	    - The TTL is extended on each access, so it's time-to-live since last access
-//
-//	Optional behavior (when calling code uses GetUnexpired() to access the cache):
-//	    - The value is only returned if it has not expired yet (regardless of whether it has been reaped)
-//	    - The TTL is NOT extended on each access
 type Manager interface {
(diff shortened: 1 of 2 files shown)


pkg/cache/cache.go
```diff
@@ -54,17 +41,36 @@ type CInterface interface {
 	IsEnabled() bool
 	Delete(key string) bool
 
-	Get(key string) interface{}
-	GetUnexpired(key string) interface{}
+	// Get a cached value by key name
+	//    - when includeExpired set to true, items existed longer than TTL will be returned, this can be useful if you cannot get the new value and need a backup. Or you need a cached without expiry.
```
peterbroadhurst, line 45:
```suggestion
	//    - when includeExpired set to true, items that have reached their TTL, but not yet been removed from the cache by any reaping logic, will be returned.
	//       This was the default behavior before the includeExpired parameter was introduced
```

pkg/cache/cache.go
```diff
@@ -54,17 +41,36 @@ type CInterface interface {
...
+	//    - when includeExpired set to false, items existed longer than TTL won't be returned.
+	//
+	// **NOTE: if you used the old Get function without the includeExpired parameter, you were getting expired items, and there wasn't a way to determine unexpired items.
```
peterbroadhurst, line 48: Think you can just remove this note

peterbroadhurst (changes requested):

pkg/cache/cache.go
```diff
@@ -82,22 +88,12 @@ func (c *CCache) Set(key string, val interface{}) {
 		c.cache.Set(key, val, c.cacheTTL)
 	}
 }
-func (c *CCache) Get(key string) interface{} {
+func (c *CCache) Get(key string, includeExpired bool) interface{} {
 	if c.enabled {
 		if cached := c.cache.Get(key); cached != nil {
-			cached.Extend(c.cacheTTL)
```
peterbroadhurst, line 88: We can't remove the extend on get functionality, as that's the function expected by 100% of the existing code paths in Core that are using this feature today (and expecting the reaper to function, which I'm sorry I've not got around to fixing yet)

Chengxuan:
replaced by https://github.com/hyperledger/firefly-common/pull/86

PR #86, opened 2023-06-29
Add behaviors to cache rather than muddying the Get function

peterbroadhurst:


.golangci.yml
```diff
@@ -32,7 +32,6 @@ linters:
   disable-all: false
   enable:
     - bodyclose
-    - depguard
```
peterbroadhurst, line 35: This was suddenly firing for me

pkg/cache/cache_test.go
```diff
@@ -127,18 +127,22 @@ func TestResetCachesForNamespace(t *testing.T) {
...
+
+	cache0.Set("test", "will not be returned after expiry")
+	cache1.Set("test", "will be returned after expiry")
+
+	for cache0.Get("test") != nil {
 		time.Sleep(1 * time.Millisecond)
 	}
 
-	assert.NotNil(t, cache1.GetUnexpired("test"))
+	assert.NotNil(t, cache1.Get("test"))
```
Chengxuan, line 147: @peterbroadhurst can we also added a test to demonstrate how TTL works for ttlFromInitialAdd mode?

Chengxuan: dumb question, not feasible to test it until reaper is added

Chengxuan (approved):

PR #87, opened 2023-07-06
Fix escaping in LIKE expressions

### LINKED ISSUE #83: Underscores in LIKE expression are not properly escaped
The current code attempts to escape `[`, `%`, and `_` by enclosing them in square brackets. I'm not sure where this originated, but it does not appear to be the correct escaping format for any of the databases currently in use. As a result, any `LIKE` query (including contains, starts with, ends with) will fail if the query value contains one of these characters.

https://github.com/hyperledger/firefly-common/blob/17b0710a0a45fdb3f997531a40ac2b47a9c0bb64/pkg/dbsql/filter_sql.go#L117-L119

In both PostgreSQL and SQLite:
* `[` does not appear to have any special meaning
* `%` and `_` need to be escaped when included in a `LIKE` expression, due to their special meaning
* Both can be escaped by prepending an escape char, which is configurable with an `ESCAPE` expression
* Postgres defaults the escape char to `\`, but SQLite does not appear to have a default

Inferring from the code here, it's possible that the intention was to use `[` as the escape character... but it would need to be explicitly specified, and there is no closing bracket needed.

Example: `LIKE "pool[_" ESCAPE "["`

Reference:
https://www.postgresql.org/docs/current/functions-matching.html
https://www.sqlite.org/lang_expr.html

awrichar:
Explicitly specify '[' as the escape character in all cases, and use it to escape the wildcards '%' and '_' in LIKE expressions. Works for Postgres and SQLite.

Fixes #83


pkg/dbsql/filter_sql.go
```diff
@@ -26,6 +26,72 @@ import (
...
+}
+
+// Convert a list of Sqlizer operations to sq.And
+func toAnd[T sq.Sqlizer](ops []T) sq.And {
+	var and sq.And
+	for _, op := range ops {
+		and = append(and, op)
+	}
+	return and
+}
```
awrichar, line 52: There might be better ways to do this, but I wanted a way to add `ESCAPE` to every `LIKE` condition without writing the iterating logic 4 times below.

Note that this logic likely isn't even used in practice, as we only construct single-element maps anyway when using `filterOp`.

pkg/dbsql/filter_sql_test.go
```diff
@@ -156,7 +156,19 @@ func TestSQLQueryFactoryEvenMoreOps(t *testing.T) {
...
+
+func TestSQLQueryFactoryEscapeLike(t *testing.T) {
+
+	sel := squirrel.Select("*").From("mytable AS mt").
+		Where(LikeEscape{"a": 1, "b": 2}).
+		Where(NotLikeEscape{"a": 1, "b": 2}).
+		Where(ILikeEscape{"a": 1, "b": 2}).
+		Where(NotILikeEscape{"a": 1, "b": 2})
+
+	_, _, err := sel.ToSql()
```
awrichar, line 170: A good test would validate more things here, but since the ordering of maps is not stable, the ordering of the generated SQL is also unpredictable...

pkg/dbsql/filter_sql.go
```diff
@@ -26,6 +26,70 @@ import (
...
+	return exprs
+}
+
+// Convert a list of Sqlizer operations to sq.And
+func toAnd[T sq.Sqlizer](ops []T) (and sq.And) {
+	for _, op := range ops {
+		and = append(and, op)
+	}
+	return and
+}
```
awrichar, line 50: There might be better ways to do this, but I wanted a way to add `ESCAPE` to every `LIKE` condition without writing the iterating logic 4 times below.

Note that this logic likely isn't even used in practice, as we only construct single-element maps anyway when using `filterOp`.

pkg/dbsql/filter_sql.go
```diff
@@ -26,6 +26,70 @@ import (
...
+
+func (lk LikeEscape) ToSql() (sql string, args []interface{}, err error) {
```
peterbroadhurst, line 52: I think letting the linter be on this one with `ToSQL` is most consistent, rather than the linter override.

peterbroadhurst (approved):
Leaving approval with one thought on letting the linter have its opinion :)

pkg/dbsql/filter_sql.go
awrichar: So... the method has to be called `ToSql` in order to implement `sq.Sqlizer`. I think that means I have to add an inline override (which I could not for the life of me figure out how to do) or a global override.

PR #89, opened 2023-07-12
config for always returning pagination response

hosie:
The structure of an array response is currently dependant on the query parameters. If `count=true` is specified, then the response is a pagination object wrapping the array and containing count and total.  Otherwise, the response is the bare JSON array.

The bare array is nice and easy to consume but it may be undesirable in certain situations to lull the developer into expecting a bare array and suddenly realising that once the collection contains so many items, that it will be paginated and the response structure suddenly changes.

This PR proposes to give the deployer of any service using firefly-common to configure it such that the pagination response structure is always used regardless of whether pagination filter was requested or whether or not the total number of records is available.

### DIFF
--- a/pkg/ffapi/apirequest.go
+++ b/pkg/ffapi/apirequest.go
@@ -1,4 +1,4 @@
-// Copyright © 2022 Kaleido, Inc.
+// Copyright © 2023 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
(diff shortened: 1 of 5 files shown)


pkg/ffapi/apirequest.go
```diff
@@ -31,17 +31,24 @@ type APIRequest struct {
...
 // FilterResult is a helper to transform a filter result into a REST API standard payload
 func (r *APIRequest) FilterResult(items interface{}, res *FilterResult, err error) (interface{}, error) {
 	itemsVal := reflect.ValueOf(items)
-	if err != nil || res == nil || res.TotalCount == nil || itemsVal.Kind() != reflect.Slice {
-		return items, err
+	if itemsVal.Kind() == reflect.Slice && (r.AlwaysPaginate || (res != nil && res.TotalCount != nil)) {
+		response := &FilterResultsWithCount{
+			Items: items,
+		}
+		if res != nil && res.TotalCount != nil {
```
peterbroadhurst, line 44: Think you might need to change the `Total` field to an `*int` pointer with `omitempty` on the `FilterResultsWithCount` - as wen need to distinguish between `"total": 0` meaning "there are zero items" and a missing total meaning "we didn't calculate one"

peterbroadhurst (changes requested):
Looks great @hosie - one hopefully small thing to close out on I thought of

PR #90, opened 2023-07-17
convert int value unset log to trace

Chengxuan:
As described in the title, I'm using the behaviour of getting the 0 big.Int back when the string is empty, it starts to become annoying to see this log line at debug level.


peterbroadhurst (approved):
👍 - agree with this

PR #91, opened 2023-07-17
Add field selection on GetMany and tweaks to Filter API builder pattern

peterbroadhurst:
- Support field-level selection/redaction in `CRUD` only
   - Adds to base `Filter` library
   - Implements in `CRUD` helper
   - Adds to API/Swagger _only if_ `SupportFieldRedaction` is set
       - This is because FireFly Core is not using `CRUD` library and hence will not support this
- Convenience of having modifiers on the base `FilterBuilder` so you can do:
    ```go
    fb.And(conditions...).Limit(1).Sort("-created") // possible before: results in Filter
    fb.Limit(1).Sort("-created").And(conditions...) // now possible: results in AndFilter
    ```

### DIFF
--- a/pkg/dbsql/crud.go
+++ b/pkg/dbsql/crud.go
@@ -434,10 +434,25 @@ func (c *CrudBase[T]) scanRow(ctx context.Context, cols []string, row *sql.Rows)
 	return inst, nil
 }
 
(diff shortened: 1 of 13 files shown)


nguyer:
Unrelated question, but at some point in the future should FireFly core move to using the `CRUD` library?

PR #92, opened 2023-07-25
Add mods to fields, specifically for "lower(field)" in SQL

peterbroadhurst:
Have a microservice with a requirement to use `lower(address)` function on query filters against the DB, in order to work with a PostgreSQL index on `lower(address)`. This allows efficient indexed query, using case insensitive searches (for example using the checksumed Eth address).

With this PR, it should be as simple as:
https://github.com/hyperledger/firefly-common/blob/8c28b1d53192521a17931bed7dfd8c68de37d5c2/pkg/dbsql/filter_sql_test.go#L43


nguyer (approved):
Looks good! It did take me a minute to figure out that `fmt.Sprintf("lower(%s)", fieldName)` was not a way for _Go_ to lowercase the string, but for the DB to lowercase the string 🙃 Need more coffee still.

PR #100, opened 2023-09-19
Define richer "ColumnsExt" syntax for CRUD base

awrichar:
Rather than continuing to add per-column flags in a piecemeal fashion, define Columns as a map. The following concepts have been migrated from top-level CRUD configs to per-column settings (although the original ones will still be honored):
* read-only columns
* immutable columns
* query modifiers

**New column features**
Each column may override its `Select` query. This allows selecting complex fields which don't actually map to a single column in this table (such as foreign columns, sub-queries, or data processed by [Postgres functions](https://www.postgresql.org/docs/9.1/functions.html)).

Each column may specify `OmitDefault` in order to be excluded from queries by default. When querying via REST, such columns can be included by using the existing `fields=` syntax to choose all fields explicitly, or a new `extrafields=` syntax to receive a few extra fields along with all the default fields.

### DIFF
--- a/pkg/dbsql/crud.go
+++ b/pkg/dbsql/crud.go
@@ -119,6 +119,7 @@ type Column[T Resource] struct {
 	Select        string
 	Immutable     bool // disallow update
 	ReadOnly      bool // disallow insert and update
+	OmitDefault   bool // omit from queries unless requested
 	QueryModifier QueryModifier
 	GetFieldPtr   func(inst T) interface{}
 }
@@ -556,6 +557,15 @@ func (c *CrudBase[T]) getCols() (cols namedColumnList[T]) {
 	return cols
 }
 
+func shouldInclude(col string, cols []string) bool {
+	for _, extra := range cols {
+		if col == extra {
+			return true
+		}
+	}
+	return false
+}
+
 func (c *CrudBase[T]) getReadCols(f *ffapi.FilterInfo) (tableFrom string, cols namedColumnList[T], readCols []string, modifiers []QueryModifier) {
 	cols = c.getCols()
 	newCols := namedColumnList[T]{cols[0] /* first column is always the sequence, and must be */}
@@ -571,8 +581,17 @@ func (c *CrudBase[T]) getReadCols(f *ffapi.FilterInfo) (tableFrom string, cols n
 				}
 			}
 		}
-		cols = newCols
+	} else {
+		for i, col := range cols {
+			if i > 0 /* idx==0 handled above */ &&
+				(!col.OmitDefault ||
+					(f != nil && shouldInclude(col.Name, f.ExtraFields))) {
+				newCols = append(newCols, col)
+			}
+		}
 	}
+	cols = newCols
+
 	tableFrom = c.Table
 	if c.ReadTableAlias != "" {
 		tableFrom = fmt.Sprintf("%s AS %s", c.Table, c.ReadTableAlias)
--- a/pkg/dbsql/crud.go
+++ b/pkg/dbsql/crud.go
@@ -112,28 +112,57 @@ type CRUD[T Resource] interface {
 	Scoped(scope sq.Eq) *CrudBase[T]                                                              // allows dynamic scoping to a collection
 }
 
+type QueryModifier func(sq.SelectBuilder) sq.SelectBuilder
+type IDValidator func(ctx context.Context, idStr string) error
+
+type Column[T Resource] struct {
+	Select        string
+	Immutable     bool // disallow update
+	ReadOnly      bool // disallow insert and update
+	QueryModifier QueryModifier
+	GetFieldPtr   func(inst T) interface{}
+}
+
+type ColumnMap[T Resource] map[string]*Column[T]
+
+type namedColumn[T Resource] struct {
+	Column[T]
+	Name string
+}
+
+type namedColumnList[T Resource] []*namedColumn[T]
+
+func (c *Column[T]) getFieldValue(inst T) interface{} {
+	// Validate() will have checked this is safe for microservices (as long as they use that at build time in their UTs)
+	return reflect.ValueOf(c.GetFieldPtr(inst)).Elem().Interface()
+}
+
 type CrudBase[T Resource] struct {
-	DB               *Database
-	Table            string
-	Columns          []string
-	FilterFieldMap   map[string]string
-	TimesDisabled    bool // no management of the time columns
-	PatchDisabled    bool // allows non-pointer fields, but prevents UpdateSparse function
-	ImmutableColumns []string
-	NameField        string                                        // If supporting name semantics
-	QueryFactory     ffapi.QueryFactory                            // Must be set when name is set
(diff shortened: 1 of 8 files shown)


pkg/dbsql/crud.go
```diff
@@ -526,19 +542,25 @@ func processGetOpts(ctx context.Context, getOpts []GetOption) (failNotFound bool
 }
 
 func (c *CrudBase[T]) GetByID(ctx context.Context, id string, getOpts ...GetOption) (inst T, err error) {
+	return c.GetByIDEnriched(ctx, id, nil, getOpts...)
+}
 
+func (c *CrudBase[T]) GetByIDEnriched(ctx context.Context, id string, enrichment []string, getOpts ...GetOption) (inst T, err error) {
```
peterbroadhurst, line 548: Could these be changed into `GetOption` constants (changing the signature of the type of `GetOption` to a function/interface), thus avoiding needing new functions for each time we add such a modifier?

pkg/dbsql/crud.go
```diff
@@ -112,6 +117,13 @@ type CRUD[T Resource] interface {
 	Scoped(scope sq.Eq) *CrudBase[T]                                                              // allows dynamic scoping to a collection
 }
 
+type QueryModifier func(sq.SelectBuilder) sq.SelectBuilder
+
+type Enrichment struct {
+	Columns       []string
```
peterbroadhurst, line 123: How do the strings here relate to the column name transformations performed between the filter specific, and the DB?

e.g. do they need to be defined in the columns of the CRUD definition?
... if so, how is this different to the specifying a limited number of columns from a subset?

Sorry for the probably quite basic questions. Maybe a test demonstrating the use would help me most.

awrichar: Mainly intended if you want to add joins, and then select columns from the joined tables. Let me see if I can put together an example usage.

pkg/dbsql/crud.go
awrichar: I thought about it... I think I would have to make `GetOption` into `interface{}` though and then parse the type of each option to see if they're an int or struct?

pkg/dbsql/crud.go
awrichar: So the answer is no, they don't need to be in the main column list. This allows you to selectively _add_ additional columns to the list, especially columns that require slightly more legwork to retrieve (so they aren't included in the default list).

pkg/dbsql/crud.go
```diff
@@ -112,28 +112,129 @@ type CRUD[T Resource] interface {
...
 
+type QueryModifier func(sq.SelectBuilder) sq.SelectBuilder
+type IDValidator func(ctx context.Context, idStr string) error
+
+type Column[T Resource] struct {
+	Select        string
+	Immutable     bool // disallow update
+	ReadOnly      bool // disallow insert and update
+	OmitDefault   bool // omit from queries unless requested
+	QueryModifier QueryModifier
```
awrichar, line 123: I'm not positive if this is the correct place for `QueryModifier`. I think it makes sense to be able to turn modifiers on and off (since they are usually `JOINs`). In this design, the use of `fields=` and `extrafields=` has the side-effect of turning on the modifiers needed only for those fields. It works very well when you have one optional field to pull in from a related table (such as enriching a linked "id" to a "name").

However, a single `JOIN` might bring in multiple extra columns... and it's a bit tricky to decide "perform this join if any of these N fields are needed", because we don't have any notion of groups of columns.

pkg/dbsql/filter_sql.go
```diff
@@ -191,6 +191,15 @@ func (s *Database) FilterUpdate(ctx context.Context, update sq.UpdateBuilder, fi
 	return update.Where(fop), nil
 }
 
+func (s *Database) aliasColumn(tableName, fieldName string) string {
+	switch {
+	case tableName == "" || strings.Contains(fieldName, "."):
```
awrichar, line 196: I don't particularly like this check for `.`... for now I copied it from `crud.go`. It seems like we may need to propagate a richer notion of tables/fields/joins throughout all these db methods.

pkg/dbsql/crud.go
```diff
@@ -112,28 +112,129 @@ type CRUD[T Resource] interface {
...
 }
 
+type QueryModifier func(sq.SelectBuilder) sq.SelectBuilder
+type IDValidator func(ctx context.Context, idStr string) error
+
+type Column[T Resource] struct {
+	Select        string
+	Immutable     bool // disallow update
+	ReadOnly      bool // disallow insert and update
+	OmitDefault   bool // omit from queries unless requested
```
awrichar, line 122: I'm open to naming on this `OmitDefault` param and its friend `ExtraFields`. I realize that these are adding some extra complexity in the name of allowing users to tailor the efficiency of the underlying database query.

pkg/dbsql/crud.go
```diff
@@ -461,40 +550,84 @@ func (c *CrudBase[T]) scanRow(ctx context.Context, cols []string, row *sql.Rows)
...
+}
+
+func (c *CrudBase[T]) legacyColumn(colName string, col *Column[T]) *namedColumn[T] {
+	col.GetFieldPtr = func(inst T) interface{} {
+		return c.GetFieldPtr(inst, colName)
+	}
+	return &namedColumn[T]{Name: colName, Column: *col}
+}
+
+func (c *CrudBase[T]) getCols() (cols namedColumnList[T]) {
```
awrichar, line 569: This method now builds the column list from both the legacy and new-style definitions. It's called somewhat frequenty, and the output should be the same every time, so I wonder if we should cache it...

pkg/dbsql/crud.go
awrichar: I also have a very slight concern that this flag doesn't get surfaced on OpenAPI. Without extra documentation, I will not know the list of default fields that will be returned, or the list of extra fields I could choose to request.

pkg/dbsql/crud.go
```diff
@@ -526,18 +659,17 @@ func processGetOpts(ctx context.Context, getOpts []GetOption) (failNotFound bool
 }
 
 func (c *CrudBase[T]) GetByID(ctx context.Context, id string, getOpts ...GetOption) (inst T, err error) {
```
awrichar, line 661: There's currently no way to pass `RequiredFields` or `ExtraFields` on any of the "get one" APIs - so no way to get anything other than the default set of fields. That feels like something that should be added as well.

pkg/dbsql/crud.go
peterbroadhurst: Yes - it's a good point, but I'm not sure I have a suggestion for how to do that easily... given the trifecta of `ff...` struct tags, `QueryField` definitions, and `CRUD` object tagging.

pkg/dbsql/crud.go
peterbroadhurst: hmmn, so really what we want is to express a join, and the column relationships for the join, and the columns you get from the join.

Then when you ask for fields, you would get them via the join.

awrichar:
I think this is now stale. We've made a few other enhancements to expose parts of the filter and crud syntax - may revisit an enhanced column definition in the future, but I think it would start from a different place.

PR #101, opened 2023-09-25
add translation missing from earlier PR

annamcallister:
https://github.com/hyperledger/firefly-common/pull/99 didn't add a translation for the new property meaning config files can't be generated.


peterbroadhurst (approved):
Looks great.

An observation that it might be good to add a build task that validates we have translations for all the built-in features.

But that's not required to merge this.

PR #106, opened 2023-10-27
Add "ModifyQuery" to CRUD interface for arbitrary SQL processing

awrichar:
While ReadQueryModifier allows CRUD models to have static additions (for joins or complex filtering), this will allow for more dynamic additions to any query.


peterbroadhurst (approved):
Looks great.

As we've discussed, a wider step back and look at a "V2" of the interface that reconciles the options a bit, would be a great thing at some point.

But this feels perfectly in-ilne with the additive nature of function we've done so far

PR #107, opened 2023-10-27
Add support for array parameters in query string

awrichar:


peterbroadhurst (approved):
So this PR changes the `Routes` object, to allow you to specify a custom query parameter for your route that accepts multiple values.

```go
		QueryParams: []*QueryParam{
			{Name: "scope", IsArray: true},
		},
```

Values marked `IsArray: true` are not passed in the `QP` field to your router function (which is limited to a single value per query parameter), but instead in the `QAP` map - which has an array of values.

> Note that we do already have one case where array values are supported, which is the built-in filter builder on the router. That has its own semantic for how `AND` or `OR` processing is used when you have multiple fields of the same name.

PR #108, opened 2023-10-27
Expose new utility ParseFilterParam

awrichar:
This takes a parameter name and value (or values) expressed in the ffapi REST query language, and converts it to an ffapi.Filter.

No functionality is changed - this is just an internal utility becoming an external one so that the query language can be used standalone.


peterbroadhurst (approved):
👍

PR #111, opened 2023-11-21
[wsclient] Allow for Embedded Basic Auth in URLs

hfuss:
We observed when using the `wsclient` package that if a WS URL had basic auth embedded in it we'd get the following error:
```
Kaleido Block IndexerCopyright (C) 2023 Kaleido
Version:  (Build Date: 2023-11-16T19:49:14Z)

time="2023-11-20T20:14:29.999Z" level=info msg="[i] No TLS configuration provided`"
time="2023-11-20T20:14:29.999Z" level=debug msg="Created REST client to https://user:pass@acme.com"
time="2023-11-20T20:14:29.999Z" level=warning msg="WS wss://user:pass@acme.com connect attempt 1 failed [-1]: "
time="2023-11-20T20:14:29.999Z" level=fatal msg="[!] Failed to connect to blockchain : FF00148: Websocket connect failed: malformed ws or wss URL"
```

In some cases, applied configurations are more difficult to change than a binary version. This change proposes that the credentials in the URL be handled more gracefully, in that 1) the embedded credentials are always removed from the URL string, and that 2) if the config did not have auth credentials provided, the embedded credentials are extracted and used automatically.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -453,10 +453,16 @@ func TestConfigWatchE2E(t *testing.T) {
 	logrus.SetLevel(logrus.DebugLevel)
 	tmpDir := t.TempDir()
 
+	// Create the file
+	os.WriteFile(fmt.Sprintf("%s/test.yaml", tmpDir), []byte(`{"ut_conf": "one"}`), 0664)
+
+	// Read initial config
 	viper.SetConfigType("yaml")
 	viper.SetConfigFile(fmt.Sprintf("%s/test.yaml", tmpDir))
+	viper.ReadInConfig()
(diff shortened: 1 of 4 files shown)


nguyer (commented):
Looks good to me. @peterbroadhurst I don't know if you want to look at the UT changes here that I made (doesn't seem right for me to approve my own changes). Essentially when running in GitHub actions, the file watcher would fail to start because the file didn't exist yet. This seems to work fine running locally, and the way the test was written it seemed like that's exactly what it was designed to test. But it doesn't work. I re-arranged the test so that the file is there when the watcher starts. But I wanted to make sure I wasn't just patching a test to cover up an actual bug.

peterbroadhurst (approved):

PR #113, opened 2023-12-03
Add JSON serialization of filters for building more complex queries

peterbroadhurst:
### Primary change

The URL query parameter spelling is designed to be convenient for simple cases, but it gets awkward with complex cases.

This PR proposes a JSON payload format and a utility to convert that into a runtime filter.

There are loads of ways this could be spelt (and many contrasting examples out there from other projects), with various compromises... the one I picked in this proposal was mainly for ease of implementation with our existing filtering code.

Some of the most significant choices/characteristics:

- Operators are the parent, values and modifiers are the children:
    - ... so `tag != 'abc'` is spelt `equal: [{not: true, field: "tag", value: "abc"}]`
- Field names are always in a `field` property - never a key name
- Operator -> array of matches semantics (makes it very easy to parse in Go strong typing)
- `AND` combination is the standard for all operator arrays, and when multiple operators used
- `OR` has to be specified explicitly, and the containing entries are all implicit `AND` (although singulars are optimized out)
### Minor other proposals in PR

- Reducing `debug` logging to one line per SQL execution
     - Apart from in the case of `CRUD.GetMany()` where we still have two, but the 2nd includes the count returned
- Relaxing the requirement on the `eventstreams` package to require IDs to be UUIDs
    - This matches then the CRUD package

### Example (from test)

```js
{
	"skip": 5,
	"limit": 10,
	"sort": [
		"tag",
		"-sequence"
	],
	"equal": [
		{
			"field": "tag",
			"value": "a"
		}
	],
	"greaterThan": [
		{
			"field": "sequence",
			"value": 10
		}
	],
	"or": [
		{
			"equal": [
				{
					"field": "masked",
					"value": true
				}
			],
			"in": [
				{
					"field": "tag",
					"values": [
						"a",
						"b",
						"c"
					]
				}
			]
		},
		{
			"equal": [
				{
					"field": "masked",
					"value": false
				}
			]
		}
	]
}
```

### The AND/OR this results in

```
(
  tag == 'a'
)
AND
(
  sequence >> 10
)
AND
(
  (
    (
      masked == true
    )
    AND
    (
      tag IN ['a','b','c']
    )
  )
  OR
  (
    masked == false
  )
)
sort=tag,-sequence skip=5 limit=10
```


pkg/ffapi/filter.go
```diff
@@ -70,6 +70,7 @@ type MultiConditionFilter interface {
 	Filter
 	// Add adds filters to the condition
 	Condition(...Filter) MultiConditionFilter
+	Conditions() []Filter
```
awrichar, line 73: Slightly confusing that `Condition` is a setter and `Conditions` is a getter...

pkg/ffapi/restfilter_json.go
```diff
@@ -0,0 +1,282 @@
...
+type FilterJSON struct {
+	Or                 []*FilterJSON          `ffstruct:"FilterJSON" json:"or,omitempty"`
+	Equal              []*FilterJSONKeyValue  `ffstruct:"FilterJSON" json:"equal,omitempty"`
+	Contains           []*FilterJSONKeyValue  `ffstruct:"FilterJSON" json:"contains,omitempty"`
+	StartsWith         []*FilterJSONKeyValue  `ffstruct:"FilterJSON" json:"startsWith,omitempty"`
+	LessThan           []*FilterJSONKeyValue  `ffstruct:"FilterJSON" json:"lessThan,omitempty"`
+	LessThanOrEqual    []*FilterJSONKeyValue  `ffstruct:"FilterJSON" json:"lessThanOrEqual,omitempty"`
+	GreaterThan        []*FilterJSONKeyValue  `ffstruct:"FilterJSON" json:"greaterThan,omitempty"`
+	GreaterThanOrEqual []*FilterJSONKeyValue  `ffstruct:"FilterJSON" json:"greaterThanOrEqual,omitempty"`
+	In                 []*FilterJSONKeyValues `ffstruct:"FilterJSON" json:"in,omitempty"`
```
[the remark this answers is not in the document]
peterbroadhurst: I added shortcuts:
- `eq`
- `neq`
- `lt`
- `nlt`
- `gt`
- `ngt`
- `lte`
- `nlte`
- `gte`
- `ngte`

pkg/ffapi/filter.go
peterbroadhurst: Renamed to `GetConditions()`

peterbroadhurst:
I've made some changes per the very useful feedback @awrichar, and this has soaked for a bit.
My suggestion is we should merge after re-review from @awrichar

pkg/ffapi/restfilter_json.go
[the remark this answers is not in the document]
awrichar: Seeing these written out, these seem redundant: `nlt` `ngt` `nlte` `ngte` (since they can all be more logically spelled as `gte` `lte` `gt` `lt`)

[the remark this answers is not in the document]
awrichar: But `neq` and `nin` seem useful

pkg/ffapi/restfilter_json.go
```diff
@@ -0,0 +1,312 @@
...
+	for _, n := range negs {
+		n.Not = true
+	}
+	return res
+}
+
+func joinInAndNin(in, nin []*FilterJSONKeyValues) []*FilterJSONKeyValues {
+	res := make([]*FilterJSONKeyValues, len(in)+len(nin))
+	copy(res, in)
+	copy(res[len(in):], nin)
```
awrichar, line 222: Does there need to be some negating here?

peterbroadhurst: Thank you 👍

awrichar (approved):
Approved with a few minor comments

pkg/ffapi/restfilter_json.go
peterbroadhurst: Added the missing test and fixed

pkg/ffapi/restfilter_json.go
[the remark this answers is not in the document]
peterbroadhurst: removing the ngts...

PR #120, opened 2024-01-21
Provide more flexibility on how EventStream specs are stored

peterbroadhurst:
I've hit some problems in having enough flexibility on the Event Streams package, and this PR proposes to solve these:
1. The `config` sub-section on the input/output object is odd
   - You configure `websockets`/`webhooks` with a sub-structure named the same name at the top level, why if you're doing something else do you have to nest it underneath a random `config` JSON
2. Removing access to the `type` field - I have some cases where ES are exposed in multiple API endpoints, with different restrictions on the type. Specifically one case where the configuration for using event streams internally, vs. externally.
3. Removing advertising `webhooks`/`websockets` fields, if those types are not actually used
4. Customizing the persistence implementation, rather coupled to 1-3 above.

Overall this approach makes the package more flexible, and while it requires a small change to consumers (per the example one), I believe all previous cases can be solved without DB migration.

Also fixes a significant issue, where the foreground context of a HTTP call is used to init a stream on `UpsertStream`.

### DIFF
--- a/pkg/eventstreams/eventstreams.go
+++ b/pkg/eventstreams/eventstreams.go
@@ -91,7 +91,7 @@ type EventStreamSpec interface {
 type EventStreamSpecFields struct {
 	Name              *string            `ffstruct:"eventstream" json:"name,omitempty"`
 	Status            *EventStreamStatus `ffstruct:"eventstream" json:"status,omitempty"`
-	InitialSequenceID *string            `ffstruct:"eventstream" json:"initialSequenceID,omitempty"`
+	InitialSequenceID *string            `ffstruct:"eventstream" json:"initialSequenceId,omitempty"`
 	TopicFilter       *string            `ffstruct:"eventstream" json:"topicFilter,omitempty"`
 
 	ErrorHandling     *ErrorHandlingType  `ffstruct:"eventstream" json:"errorHandling"`
--- a/pkg/eventstreams/eventstreams.go
+++ b/pkg/eventstreams/eventstreams.go
@@ -365,7 +365,7 @@ func (es *eventStream[CT, DT]) checkSetStatus(ctx context.Context, targetStatus 
 }
 
 func (es *eventStream[CT, DT]) persistStatus(ctx context.Context, targetStatus EventStreamStatus) error {
-	fb := EventStreamFilters.NewUpdate(ctx)
+	fb := GenericEventStreamFilters.NewUpdate(ctx)
 	return es.esm.persistence.EventStreams().Update(ctx, es.spec.GetID(), fb.Set("status", targetStatus))
 }
 
--- a/pkg/eventstreams/eventstreams.go
+++ b/pkg/eventstreams/eventstreams.go
@@ -236,24 +236,24 @@ type EventStreamActions[CT any] interface {
 }
 
 func (esm *esManager[CT, DT]) initEventStream(
-	bgCtx context.Context,
 	spec CT,
 ) (es *eventStream[CT, DT], err error) {
 	// Validate
-	factory, err := esm.validateStream(bgCtx, spec, LifecyclePhaseStarting)
+	factory, err := esm.validateStream(esm.bgCtx, spec, LifecyclePhaseStarting)
 	if err != nil {
 		return nil, err
 	}
 
+	streamCtx := log.WithLogField(esm.bgCtx, "eventstream", *spec.ESFields().Name)
 	es = &eventStream[CT, DT]{
-		bgCtx:       log.WithLogField(bgCtx, "eventstream", *spec.ESFields().Name),
+		bgCtx:       streamCtx,
 		esm:         esm,
 		spec:        spec,
 		persistence: esm.persistence,
 		retry:       esm.config.Retry,
 	}
 
-	es.action = factory.NewDispatcher(es.bgCtx, &esm.config, spec)
+	es.action = factory.NewDispatcher(streamCtx, &esm.config, spec)
 
 	log.L(es.bgCtx).Infof("Initialized Event Stream")
 	if *spec.ESFields().Status == EventStreamStatusStarted {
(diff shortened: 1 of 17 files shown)


pkg/eventstreams/persistence.go
```diff
@@ -52,8 +53,64 @@ var CheckpointFilters = &ffapi.QueryFields{
 
 type IDValidator func(ctx context.Context, idStr string) error
 
-func NewEventStreamPersistence[CT any](db *dbsql.Database, idValidator IDValidator) Persistence[CT] {
-	return &esPersistence[CT]{
+type GenericEventStream struct {
```
peterbroadhurst, line 56: This is probably the most meaningful starting point for understanding the change.

Before: The had an `EventStreamSpec[CT any]` that was a `struct` and you **must** use it, and put any special config you have under a `config` subsection.

After: This `GenericEventStream` is a `struct` that you can use if you don't have any special config, or you can copy+extend it to add in the config you need.

This was the best compromise I could find that gave the flexibility, without making massive changes.

pkg/eventstreams/persistence.go
```diff
@@ -68,8 +125,8 @@ func (p *esPersistence[CT]) IDValidator() IDValidator {
 	return p.idValidator
 }
 
-func (p *esPersistence[CT]) EventStreams() dbsql.CRUD[*EventStreamSpec[CT]] {
-	return &dbsql.CrudBase[*EventStreamSpec[CT]]{
+func (p *esPersistence[CT]) EventStreams() dbsql.CRUD[*GenericEventStream] {
```
peterbroadhurst, line 128: This too has gone from being a mandated implementation, to a reference for you to extend.

peterbroadhurst: ... which matches up with the fact that the DB migrations are described as reference already, so overall I feel it's much more consistent this way

test/es_demo_migrations/000001_create_eventstreams_table.up.sql
```diff
@@ -9,7 +9,6 @@ CREATE TABLE eventstreams (
   initial_sequence_id  TEXT,
   topic_filter         TEXT,
   identity             TEXT,
-  config               TEXT,
```
peterbroadhurst, line 12: You are no longer constrained to this approach. If you want first-class indexed columns for your extra config, you can do it. However, you have to take ownership of the whole CRUD implementation in that case (rather than using `NewGenericEventStreamPersistence`).

pkg/eventstreams/eventstreams.go
```diff
@@ -77,45 +78,31 @@ type DBSerializable interface {
...
+	IsNil() bool                     // needed as quirk of using interfaces with generics
+}
+
+type EventStreamSpecFields struct {
 	Name              *string            `ffstruct:"eventstream" json:"name,omitempty"`
 	Status            *EventStreamStatus `ffstruct:"eventstream" json:"status,omitempty"`
-	Type              *EventStreamType   `ffstruct:"eventstream" json:"type,omitempty" ffenum:"estype"`
 	InitialSequenceID *string            `ffstruct:"eventstream" json:"initialSequenceID,omitempty"`
 	TopicFilter       *string            `ffstruct:"eventstream" json:"topicFilter,omitempty"`
-	Identity          *string            `ffstruct:"eventstream" json:"identity,omitempty"`
```
peterbroadhurst, line 89: This was an example of a field that needed to be top-level for one implementation for DB indexing, and with the previous structure that meant pushing it down to this common layer.

Now we've addressed that form of extensibility properly, it can go.

pkg/eventstreams/persistence.go
```diff
@@ -22,25 +22,26 @@ import (
...
-	"identity":    &ffapi.StringField{},
-	"config":      &ffapi.JSONField{},
+	"id":                &ffapi.StringField{},
+	"created":           &ffapi.TimeField{},
+	"updated":           &ffapi.TimeField{},
+	"name":              &ffapi.StringField{},
+	"status":            &ffapi.StringField{},
+	"type":              &ffapi.StringField{},
+	"topicfilter":       &ffapi.StringField{},
+	"identity":          &ffapi.StringField{},
```
awrichar, line 43: is this still filterable?

pkg/eventstreams/persistence.go
```diff
@@ -92,16 +149,17 @@ func (p *esPersistence[CT]) EventStreams() dbsql.CRUD[*EventStreamSpec[CT]] {
 			"websocket_config",
```
awrichar, line 149: I'm trying to wrap my head around the migration impact to anything using this before the change. What was stored in "config" vs "webhook_config" and "websocket_config"? Will migration involve extending this generic implementation to continue storing data that was in the removed columns "identity" and "config"?

pkg/eventstreams/persistence.go
peterbroadhurst: Good spot, thanks. This should be removed.
An implementation that has extra fields, has to create their own filter list.

pkg/eventstreams/persistence.go
peterbroadhurst: Yes - exactly.
You can still choose to store that configuration in DB fields called `identity` and `config`, but you're no longer forced to.

If you were doing so before, you would indeed need to move to using your own `Persistence` implementation - in fact this library is very opinionated now that if you do want to extend you _should_ take on the persistence implementation yourself.

peterbroadhurst: I understand it's quite a big change to the interface, where the migration impact is on consumers who have extended it.

The arguments really for the change are:
1. It's a rather new feature in the library, and I don't have evidence it's used widely yet, so it's a good time to address fundamental issues.
2. As someone trying to use the extensibility before, I found it was extremely difficult to extend previously. Now it's much easier. For me the code migration on the consuming side (about a 1/2 day) was well, well worth it in terms of where it left me for future extensibility.

pkg/eventstreams/persistence.go
peterbroadhurst: Also adding some comments to the code to clear up the role of `GenericEventStream` and these base filters.

peterbroadhurst: Ok - commit added now with that

pkg/eventstreams/persistence.go
awrichar: OK, makes sense - and the comments added in the code are very helpful to that effect. I agree we're trying to feel out the best way to provide an extensible implementation, and it's probably best to make these changes now.

awrichar (approved):

PR #121, opened 2024-01-22
fix: delete stream with name

EnriqueL8:
When deleting we add an IDFilter so that the SQL query results in `DELETE FROM table WHERE ID = <value>` in the case of deleting stream we were passing the nameOrID to that SQL query and it was failing when we passed the name. Since we retrieve the value from the DB before deleting we can just pass in the ID to delete


peterbroadhurst (approved):
Change looks great thanks @EnriqueL8 - one small build issue to address before merge.

peterbroadhurst (approved):

PR #122, opened 2024-01-24
Remove WebhookConfig.Headers in favour of WebhookConfig.HTTP.HTTPHeaders

chrisbygrave:
WebhookConfig was recently introduced with a top-level `headers` property. However it also contains an `http` property - a standard `ffrestry.HTTPConfig` object which already provides the facility for defining HTTP headers.

This PR removes the top-level `headers` property, and updates the web hook action to use `HTTP.HTTPHeaders` on the request instead.


pkg/eventstreams/webhooks.go
```diff
@@ -34,7 +34,6 @@ import (
 type WebhookConfig struct {
 	URL           *string             `ffstruct:"whconfig" json:"url,omitempty"`
 	Method        *string             `ffstruct:"whconfig" json:"method,omitempty"`
-	Headers       map[string]string   `ffstruct:"whconfig" json:"headers,omitempty"`
```
EnriqueL8, line 37: This would have to feed its way into a release note as it's a breaking change, I wonder if we want a deprecated field instead?

pkg/eventstreams/webhooks.go
```diff
@@ -115,8 +114,13 @@ func (w *webhookAction[CT, DT]) AttemptDispatch(ctx context.Context, attempt int
...
 	req.Header.Set("Content-Type", "application/json")
-	for h, v := range w.spec.Headers {
-		req.Header.Set(h, v)
+
+	if w.spec.HTTP != nil {
+		for h, v := range w.spec.HTTP.HTTPHeaders {
+			if vs, ok := v.(string); ok {
+				req.Header.Set(h, vs)
+			}
+		}
```
EnriqueL8, line 123: I think we just remove this. They get added as part of the init of the HTTP Config so this will result in double work.

EnriqueL8: https://github.com/hyperledger/firefly-common/blob/main/pkg/ffresty/ffresty.go#L211-L215

chrisbygrave: Good point - thanks

pkg/eventstreams/webhooks.go
peterbroadhurst: I suggest at the stage of lifeycle we're at with this function we're ok with the release notes item.
I can include that cutting a release on the common library.

PR #123, opened 2024-01-27
Refactor WebSockets to fix window of duplicate delivery after reconect

peterbroadhurst:
The `WebSocketChannels` interface has been around a long time, all the way back to the early days of EthConnect.

It provides a protocol for:
- Multiplexing multiple `streams` onto a single WebSocket
- Providing acknowledged at-least-once delivery of batches across multiple WebSocket connections on a single stream
   - With workload balancing of deliveries across multiple connections to the server
- Broadcasting a copy of a message to every websocket (without ack)

But there is a problem with the complex way that the channels are used as a code interface, per this function (which is derived from similar code from EthConnect, FFTM/EVMConnect and FireFly core):
https://github.com/hyperledger/firefly-common/blob/badf0651f1b7d98b9a0b2367e2625d4556f74e70/pkg/eventstreams/websockets.go#L78-L112

Because after a disconnect when messages arrive that are candidates for delivery, a `send` wiil be queued into the channel when there are no WebSockets connected to receive it. Then when a WebSocket connects, that websocket will receive the payload/batch queued into the channel for it. However, the `AttemptDispatch` will immediately fail in `waitForAck` (without waiting for an ack at all) because the WebSocket ID has changed, even thought the disconnection happened some time ago.

This does not affect the resilience of the idempotent delivery interface between FFTM and FireFly, but with the new `eventstreams` interface in microservices it is highly likely to occur and be frustrating to applications consuming the stream.

So this PR proposes to fix the problem by simplifying the API interface of the `wsserver` package (and resulting in a re-write of the stream selection and dispatch logic of `wsserver`).

The code used to be a complex pull model, where each webscoket connection tracks the streams it's started on, and pulls from a separate channel for each stream. This required uncommon code constructs, like dynamic `select` against an array of potential channels.
This PR proposes a much more straight forward model where a `RoundTrip` object is defined and the parent server-level code keeps track of the connections available for each stream - so it can pick a different one for each round-trip. It then pushes that send into a unique send channel for that webscoket. The parent is responsible for making sure each stream can have just one round-trip in-flight at any given time, which is cancelled if the in-bound WebSocket processing it disconnects.

After the change, the consumer code that uses the `wsserver` API looks like:
https://github.com/hyperledger/firefly-common/blob/07fafeebabc73f665f494ff4fbb96b8f85b0d0b9/pkg/eventstreams/websockets.go#L77-L95

The problem being solve is illustrated in the below log snippet.

```
[2024-01-26T18:42:18.650Z] TRACE Received: {Type:start Stream:sub1 Message: BatchNumber:1} pid=1 wsc=4a00e7ae-5c9b-4066-8a2c-5d7e01e71201
[2024-01-26T18:42:18.651Z]  INFO Batch 2 dispatched (len=1,attempt=0) eventstream=sub1 pid=1
[2024-01-26T18:42:18.651Z]  INFO WebSocket event batch 2 delivery failed (len=1,attempt=0): FF00228: WebSocket 'a87c0ea3-905d-4dcf-afcf-e739707c206c' closed eventstream=sub1 pid=1
[2024-01-26T18:42:18.651Z] ERROR Batch 2 attempt 0 failed. err=FF00228: WebSocket 'a87c0ea3-905d-4dcf-afcf-e739707c206c' closed eventstream=sub1 pid=1
[2024-01-26T18:42:18.651Z] ERROR action attempt 1: FF00228: WebSocket 'a87c0ea3-905d-4dcf-afcf-e739707c206c' closed eventstream=sub1 pid=1
[2024-01-26T18:42:18.901Z] DEBUG Batch 2 attempt 1 dispatching. Len=1 eventstream=sub1 pid=1
[2024-01-26T18:42:18.901Z]  INFO Batch 2 dispatched (len=1,attempt=1) eventstream=sub1 pid=1
[2024-01-26T18:42:18.901Z]  INFO WebSocket event batch 2 delivery failed (len=1,attempt=1): FF00228: WebSocket 'a87c0ea3-905d-4dcf-afcf-e739707c206c' closed eventstream=sub1 pid=1
[2024-01-26T18:42:18.901Z] ERROR Batch 2 attempt 1 failed. err=FF00228: WebSocket 'a87c0ea3-905d-4dcf-afcf-e739707c206c' closed eventstream=sub1 pid=1
[2024-01-26T18:42:18.901Z] ERROR action attempt 2: FF00228: WebSocket 'a87c0ea3-905d-4dcf-afcf-e739707c206c' closed eventstream=sub1 pid=1
[2024-01-26T18:42:19.402Z] DEBUG Batch 2 attempt 2 dispatching. Len=1 eventstream=sub1 pid=1
[2024-01-26T18:42:19.402Z]  INFO Batch 2 dispatched (len=1,attempt=2) eventstream=sub1 pid=1
```

### Additional change - extensibility for JSON based query

Sorry @awrichar, I meant to split this out to a separate PR.
I've added an extra interface to the `ffapi.QueryJSON` interface, to allow extensibility for a use case where some of the fields are names that need to be resolved to ID before the query runs.

The interface gives you access to the whole query tree at the level you are at, which in my use case is because you might need to look at (/require) extra `equal: [...]` options at the same level in the tree to scope the name lookup.

This extensibility is only added for the complex JSON format of query, because the problem is much simpler in the path + query-field case and I'm not sure there's an obvious way to join the logic (without a big restructure).


pkg/wsserver/wsserver.go
```diff
@@ -91,25 +103,24 @@ func (s *webSocketServer) Handler(w http.ResponseWriter, r *http.Request) {
...
-	defer s.mux.Unlock()
-
-	// When a connection that was listening on a stream closes, we need to wake anyone
-	// that was listening for a response
-	select {
-	case t.receiverChannel <- &WebSocketCommandMessageOrError{Err: i18n.NewError(s.ctx, i18n.MsgWebSocketClosed, connInfo)}:
-	default:
-	}
-}
+// i18n.NewError(s.ctx, i18n.MsgWebSocketClosed, connInfo)
```
EnriqueL8, line 106: Do you need this?

EnriqueL8 (commented):
Amazing stuff!

pkg/wsserver/wsserver.go
```diff
@@ -28,54 +27,67 @@ import (
...
-	WebSocketChannels
+	Protocol
 	Handler(w http.ResponseWriter, r *http.Request)
 	Close()
 }
 
+type streamState struct {
+	wlmCounter int64
+	inflight   *roundTrip
+	conns      []*webSocketConnection
```
EnriqueL8, line 65: might be useful making this a map with the id being the key as we iterate over this a few times to delete and add

peterbroadhurst: this actually changed _away_ from a map in this PR, because the performance sensitive part is a copy of the list that has to happen on every broadcast. That can be a simple mem-copy now.

Spot on that the other side of the coin is the need to iterate on the lifecycle action of close.

Let me know what you think after considering, as it was a big thinking point as I worked on it.

EnriqueL8: I see - so we copy so that we don't retain a mutex.Lock in case a new connection is added or deleted whilst we are in the middle of a broadcast. Interesting to me that even with a `copy()` the channels in each ws connection still work, tested in Go Playground. Happy for this to be a copy instead of a map 😃

pkg/wsserver/wsserver.go
```diff
@@ -118,92 +132,155 @@ func (s *webSocketServer) Close() {
...
+		// and always handle the cleanup of the RoundTripper
+		if ss.inflight != nil {
+			return i18n.NewError(s.ctx, i18n.MsgWebSocketBatchInflight, stream, ss.inflight.batchNumber, ss.inflight.conn.id)
+		}
+		// Do a round-robbin pick of one of the connections
+		conn := ss.conns[ss.wlmCounter%int64(len(ss.conns))]
+		ss.wlmCounter++
+		// The wlmCounter is used as the batch number in the payload
+		payload.GetBatchHeader().BatchNumber = ss.wlmCounter
+		payload.GetBatchHeader().Stream = stream
```
EnriqueL8, line 173: nit: seems odd to me that it's a Get and we update in place

peterbroadhurst: This is because Go doesn't have an object model to allow something like class inheritance (which would be available in Java), so when combining with Generics this is the best patterned I've found for mandating embedding of a sub-structure into a parent.

Really `GetBatchHeader()` is `GetBatchHeaderPtr()` - but that felt unnecessary.

pkg/wsserver/config_test.go
```diff
@@ -0,0 +1,40 @@
+// Copyright © 2023 Kaleido, Inc.
```
nguyer, line 1: I guess the linter doesn't check test files 🙂

pkg/wsserver/wsconn.go
```diff
@@ -56,81 +48,44 @@ type WebSocketCommandMessage struct {
...
-		c.server.cycleStream(c.id, t)
-		log.L(c.ctx).Infof("Websocket closed while active on stream '%s'", t.streamName)
-	}
 	c.server.connectionClosed(c)
 	log.L(c.ctx).Infof("Disconnected")
 }
 
 func (c *webSocketConnection) sender() {
 	defer c.close()
-	buildCases := func() []reflect.SelectCase {
```
nguyer, line 92: I am so happy to see this go! 👋

nguyer (approved):

PR #124, opened 2024-01-29
fix: Change how we index the filter operations before evaluation

SamMayWork:
Crashes have been observed when using query filters that have a single `eq` operator in them, this changes the logic so that we actually index the array correctly for these kinds of filters.

(Also adds tests for other short-name operations to prove that those work as intended too.)

### DIFF
--- a/pkg/ffapi/restfilter_json.go
+++ b/pkg/ffapi/restfilter_json.go
@@ -210,13 +210,7 @@ func (jq *QueryJSON) addSimpleFilters(ctx context.Context, fb FilterBuilder, jso
 func joinShortNames(long, short, negated []*FilterJSONKeyValue) []*FilterJSONKeyValue {
 	res := make([]*FilterJSONKeyValue, len(long)+len(short)+len(negated))
 	copy(res, long)
-
-	if len(long) == 0 {
-		copy(res[0:], short)
-	} else {
-		copy(res[len(long):], short)
-	}
-
+	copy(res[len(long):], short)
 	negs := res[len(short)+len(long):]
 	copy(negs, negated)
 	for _, n := range negs {
--- a/pkg/ffapi/restfilter_json.go
+++ b/pkg/ffapi/restfilter_json.go
@@ -210,7 +210,13 @@ func (jq *QueryJSON) addSimpleFilters(ctx context.Context, fb FilterBuilder, jso
 func joinShortNames(long, short, negated []*FilterJSONKeyValue) []*FilterJSONKeyValue {
 	res := make([]*FilterJSONKeyValue, len(long)+len(short)+len(negated))
 	copy(res, long)
-	copy(res[len(short):], short)
+
+	if len(long) == 0 {
+		copy(res[0:], short)
+	} else {
+		copy(res[len(short):], short)
+	}
+
 	negs := res[len(short)+len(long):]
 	copy(negs, negated)
 	for _, n := range negs {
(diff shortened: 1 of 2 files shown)


pkg/ffapi/restfilter_json.go
```diff
@@ -210,7 +210,13 @@ func (jq *QueryJSON) addSimpleFilters(ctx context.Context, fb FilterBuilder, jso
 func joinShortNames(long, short, negated []*FilterJSONKeyValue) []*FilterJSONKeyValue {
 	res := make([]*FilterJSONKeyValue, len(long)+len(short)+len(negated))
 	copy(res, long)
-	copy(res[len(short):], short)
+
+	if len(long) == 0 {
+		copy(res[0:], short)
+	} else {
+		copy(res[len(short):], short)
```
EnriqueL8, line 217: Shouldn't this be `len(long)` or `len(res)`? Since we want to copy into from index of where long left off to short... 
```suggestion
		copy(res[len(res):], short)
```

EnriqueL8 (commented):
Great find - This function is quiet messy, personally I would refactor it do perform appends. `append(array1, array2...)` I know probably `copy` is faster but we are dealing with indexes it can be confusing. There might be a reason with want to do shallow copy as we are modifying it in the for loop for negs?

pkg/ffapi/restfilter_json.go
SamMayWork: This absolutely should be either `long` or `res` here (my bad) and the UT should have caught it but didn't, so will change that too.

This [comment](https://stackoverflow.com/a/58726780) indicates some potentially nasty behaviour with `append` and the existing specific structure here makes me worry for unintended consequences🔥

SamMayWork: Resolved in https://github.com/hyperledger/firefly-common/pull/124/commits/d549fecace2166b1738d25f90a9cd2ee97d6fe47, looking further at the code `short` as the index value is obviously wrong, but would actually never cause an issue since if you provide more than one `eq` value in a block, the other values are ignored. (This means the failure can't be recreated in a UT, since it's not actually a possible condition).

pkg/ffapi/restfilter_json.go
```diff
@@ -210,7 +210,13 @@ func (jq *QueryJSON) addSimpleFilters(ctx context.Context, fb FilterBuilder, jso
 func joinShortNames(long, short, negated []*FilterJSONKeyValue) []*FilterJSONKeyValue {
 	res := make([]*FilterJSONKeyValue, len(long)+len(short)+len(negated))
 	copy(res, long)
-	copy(res[len(short):], short)
```
peterbroadhurst, line 213:
```suggestion
	copy(res[len(long):], short)
```

Isn't this the bug @SamMayWork ?

peterbroadhurst: (if there were a problem with `copy` having `nil` as the right parameter, then that would be a problem in the line above, which this PR doesn't change)

SamMayWork: Yep after making the change from `short -> long` the surrounding if block is actually redundant now, removing now 🖊️

SamMayWork: Resolved in https://github.com/hyperledger/firefly-common/pull/124/commits/066994cbb94c663b8f892f44f9c1bd248db13db9

peterbroadhurst (approved):

PR #126, opened 2024-02-08
Do not lose calling context in the logger

peterbroadhurst:
The log lines that contain the `breq` are currently missing the `req` and `httpreq` context, as well as any other calling context on that. It looks like a straight typo, that we're overwriting the `rCtx` with the constructor context of the client.
```
[2024-02-08T18:43:49.638Z] DEBUG EVMConnectorBody: {....} httpreq=dJAh7YSk pid=1 req=7hvX9eW_
[2024-02-08T18:43:49.638Z] DEBUG ==> POST https:/.../ breq=6yjJwWIC pid=1 proto=ethereum
```


nguyer (approved):
The changes look good, but it looks like a commit from another branch (which was just merged) got added on here and is creating a conflict in the error codes now.

PR #127, opened 2024-02-13
[ffapi] Support for Text and Binary Streams

hfuss:
- For routing / server - adds a new `StreamHandler` extension similar to the `JSONHandler` but with the expectation that an `io.ReadCloser` is returned. The route will be assumed to be streaming binary unless a `StreamOutputContentType` is specified.
- For OpenAPI generation - adds a new `OutputType` to determine if the default JSON helpers are used, or the stream helpers. Streams are assumed to always return `200 OK` with the `Content-Type` being customizable through the same `StreamOutputContentType`.


hfuss:
Example route:
```golang
var getLogs = &ffapi.Route{
	Name:                    "getStream",
	Path:                    "someStream",
	Method:                  http.MethodGet,
	Description:             "a stream API",
	OutputType:              ffapi.RouteOutputTypeStream,
	StreamOutputContentType: "text/plain",
	PathParams:              []*ffapi.PathParam{},
	QueryParams:             []*ffapi.QueryParam{},
	JSONInputValue:          nil,

	Extensions: &ffapi.APIServerRouteExt[LogManagerRequest]{
		StreamHandler: func(r *ffapi.APIRequest, lmr LogManagerRequest) (output io.ReadCloser, err error) {
			ctx := r.Req.Context()
			log.L(ctx).Infof("in stream API handler")

			s := strings.NewReader("hello!")
			output = io.NopCloser(s)
			return
		},
	},
}
```

hfuss:
Example:
```golang
var getSomeStream = &ffapi.Route{
	Name:                    "getSomeStream",
	Path:                    "someStream",
	Method:                  http.MethodGet,
	Description:             "an API for some stream",
	OutputType:              ffapi.RouteOutputTypeStream,
	StreamOutputContentType: "text/plain",
	PathParams:              []*ffapi.PathParam{},
	QueryParams:             []*ffapi.QueryParam{},
	JSONInputValue:          nil,

	Extensions: &ffapi.APIServerRouteExt[MicroserviceRequest]{
		StreamHandler: func(r *ffapi.APIRequest, mr MicroserviceRequest) (output io.ReadCloser, err error) {
			ctx := r.Req.Context()
			log.L(ctx).Infof("in stream API handler")

			s := strings.NewReader("hello!")
			output = io.NopCloser(s)
			return
		},
	},
}
```

pkg/ffapi/openapi3.go
```diff
@@ -277,9 +277,23 @@ func CheckObjectDocumented(example interface{}) {
 }
 
 func (sg *SwaggerGen) addOutput(ctx context.Context, doc *openapi3.T, route *Route, op *openapi3.Operation) {
+	s := i18n.Expand(ctx, i18n.APISuccessResponse)
+	if route.OutputType == RouteOutputTypeStream {
```
hfuss, line 281: I thought about just checking if `route.StreamHandler` was set, but that didn't seem right bc like `JSONHandler`, the framework actually sets `StreamHandler` with a wrapper around `extensions.StreamHandler` by default. So it wasn't guaranteed to be set at this point, and without adding a generic I can't cast and inspect `Extensions` here.

pkg/ffapi/apiserver.go
```diff
@@ -201,13 +203,24 @@ func (as *apiServer[T]) routeHandler(hf *HandlerFactory, route *Route) http.Hand
 	// We extend the base ffapi functionality, with standardized DB filter support for all core resources.
 	// We also pass the Orchestrator context through
 	ext := route.Extensions.(*APIServerRouteExt[T])
-	route.JSONHandler = func(r *APIRequest) (output interface{}, err error) {
-		er, err := as.EnrichRequest(r)
-		if err != nil {
-			return nil, err
+	if route.OutputType == RouteOutputTypeStream && ext.StreamHandler != nil {
```
nguyer, line 206: Does a `switch` between available `OutputType`s make sense here?

nguyer (commented):
This makes sense to me. No objections here.

pkg/ffapi/handler.go
```diff
@@ -215,12 +215,16 @@ func (hs *HandlerFactory) RouteHandler(route *Route) http.HandlerFunc {
...
+			case multipart != nil:
 				r.FP = multipart.formParams
 				r.Part = multipart.part
-				output, err = route.FormUploadHandler(r)
-			} else {
-				output, err = route.JSONHandler(r)
+				output.out, err = route.FormUploadHandler(r)
+			case route.OutputType == RouteOutputTypeStream && route.StreamHandler != nil:
+				output.out, err = route.StreamHandler(r)
+				output.contentType = route.StreamOutputContentType
```
peterbroadhurst, line 225: I wonder about making this a dynamic return alongside the stream, or otherwise a `r.ContentType` way of modifying it (like `r.SuccessStatusCode`).

Making it static feels wrong... what if I want to return `text/csv` or `text/json` data depending on the `Accept` header in the request?

pkg/ffapi/openapi3.go
peterbroadhurst: The other option here, is just to allow `route.CustomOutput` or similar, to be the OpenAPI3 map of status codes to openapi3.ResponseRef objects.

pkg/ffapi/handler.go
hfuss: Makes perfect sense!

hfuss: After closer looking at whats in `APIRequest` I determined in my handler I should just be setting the `Content-Type` in my handler via the `ResponseHeaders`.

I then updated the handler code so that it respects the `Content-Type` if its already set. That way, `StreamHandlers` can look at the `Accept` header of the request and determine what encoding they need to make the bytes of the `io.ReadCloser` use and set their `Content-Type` accordingly.

hfuss:
Updated example after feedback:
```golang
var getSomeStream = &ffapi.Route{
	Name:                    "getSomeStream",
	Path:                    "someStream",
	Method:                  http.MethodGet,
	Description:             "an API for some stream",
	PathParams:              []*ffapi.PathParam{},
	QueryParams:             []*ffapi.QueryParam{},
	JSONInputValue:          nil,
	
	CustomResponseRefs: map[string]*openapi3.ResponseRef{
		"200": {
			Value: &openapi3.Response{
				Content: openapi3.Content{
					"text/plain": &openapi3.MediaType{},
				},
			},
		},
	},
	Extensions: &ffapi.APIServerRouteExt[MicroserviceRequest]{
		StreamHandler: func(r *ffapi.APIRequest, mr MicroserviceRequest) (output io.ReadCloser, err error) {
			ctx := r.Req.Context()
			log.L(ctx).Infof("in stream API handler")

			s := strings.NewReader("hello!")
			output = io.NopCloser(s)
			r.ResponseHeaders.Add("Content-Type", "text/plain")
			return
		},
	},
}
```

pkg/ffapi/handler.go
hfuss: Does beg the question if `StreamHandler` should just become a `RawHandler` and just return `interface{}` in case theres ever use cases for non-JSON and non-stream APIs... thoughts @nguyer @peterbroadhurst ?

nguyer (approved):

PR #128, opened 2024-02-13
Add ability to selectively ignore adding a route into the OpenAPI document

EnriqueL8:
Allow at the Route level to choose to add it to the OpenAPI generated or not


EnriqueL8:
I've realised that I can selectively choose which routes can be added to the server and the OpenAPI package. I can simply control it at that level and I don't see a scenario where I want to hide it from the OpenAPI document and not the server, so closing this PR.

PR #129, opened 2024-02-19
fix: empty stream state on broadcast

EnriqueL8:
When a broadcast is attempted and the steam state hasn't been initialise it caused a nil pointer.

Two ways to solve this:
- If the stream state is nil, set an empty one as part of broadcast but that means no connections will receive those messages until that map updates 
- Wait for a least one connection on this stream to start and then start broadcasting. At least one connection will receive the messages for the stream whilst it's connected

I've gone for the latter one. We perform best-effort delivery for broadcast, so it could be argued to do the former, any thoughts appreciated!

### DIFF
--- a/pkg/wsserver/wsserver.go
+++ b/pkg/wsserver/wsserver.go
@@ -136,15 +136,8 @@ func (s *webSocketServer) Broadcast(ctx context.Context, stream string, payload 
 	s.mux.Lock()
 	ss := s.streamMap[stream]
 	if ss == nil {
-		s.mux.Unlock()
-		// Waiting for a least one connection to start the stream
-		// before we start broadcasting
-		// Err always nil per callback function
-		_ = s.waitStreamConnections(ctx, stream, func(providedState *streamState) error {
-			ss = providedState
-			return nil
-		})
-		s.mux.Lock()
+		ss = &streamState{}
+		s.streamMap[stream] = ss
 	}
 	conns := make([]*webSocketConnection, len(ss.conns))
 	copy(conns, ss.conns)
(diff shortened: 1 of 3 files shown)


pkg/wsserver/wsserver.go
```diff
@@ -132,9 +132,20 @@ func (s *webSocketServer) Close() {
...
-func (s *webSocketServer) Broadcast(ctx context.Context, stream string, payload interface{}) {
+func (s *webSocketServer) Broadcast(ctx context.Context, stream string, payload interface{}) error {
 	s.mux.Lock()
 	ss := s.streamMap[stream]
+	if ss == nil {
+		s.mux.Unlock()
+		// Waiting for a least one connection to start the stream
+		// before we start broadcasting
+		// Err always nil per callback function
+		_ = s.waitStreamConnections(ctx, stream, func(providedState *streamState) error {
```
nguyer, line 143: Does `waitStreamConnections` block here if nobody is connected?

nguyer (commented):
I'm not sure that this solution is consistent with the way we have viewed these types of connections throughout our architecture. The "Broadcast" function is expected to be non-blocking and any code should be able to call it regardless of whether any clients are current connected. The Broadcast function specifically is also not intended to be a durable message queue (as opposed to the RoundTrip function), which I think these changes have begun to alter. It is expected that if something is broadcasted (a TX receipt for example) while my client is not connected, if I am interested in that receipt, when I need to query for it when my client connects. I expect to be caught up on batches that I have missed (the durable part of the stream), but broadcasts are a different thing altogether.

pkg/wsserver/wsserver.go
EnriqueL8: Yes it does block

nguyer:
> This will not be true I expect to be caught up on batches that I have missed and continue to be true if you are late joiner to a broadcast stream

That's fine. Broadcasts and batches that go through the RoundTrip are different in this regard.

EnriqueL8:
Just to clarify we are broadcasting batches as well so batching and delivery methods are completely different and configured independently. 

Have pushed the simpler change

nguyer (approved):

PR #132, opened 2024-03-11
[fftls] Allow for In-Memory or Inlined TLS Certificates

hfuss:
Useful for configs which want to inline the certs as large PEM blocks in their config files, or when constructing TLS configs from in-memory certificates (such as certificates downloaded from a Kubernetes Secret).

### DIFF
--- a/pkg/fftls/config.go
+++ b/pkg/fftls/config.go
@@ -1,4 +1,4 @@
-// Copyright © 2023 Kaleido, Inc.
+// Copyright © 2024 Kaleido, Inc.
 //
 // SPDX-License-Identifier: Apache-2.0
 //
--- a/pkg/fftls/config.go
+++ b/pkg/fftls/config.go
@@ -23,14 +23,21 @@ import (
 const (
 	// HTTPConfTLSCAFile the TLS certificate authority file for the HTTP server
 	HTTPConfTLSCAFile = "caFile"
+	// HTTPConfTLSCA the TLS certificate authority in PEM format
+	HTTPConfTLSCA = "ca"
 	// HTTPConfTLSCertFile the TLS certificate file for the HTTP server
 	HTTPConfTLSCertFile = "certFile"
+	// HTTPConfTLSCert the TLS certificate in PEM format
+	HTTPConfTLSCert = "cert"
 	// HTTPConfTLSClientAuth whether the HTTP server requires a mutual TLS connection
 	HTTPConfTLSClientAuth = "clientAuth"
 	// HTTPConfTLSEnabled whether TLS is enabled for the HTTP server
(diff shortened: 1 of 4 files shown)


hfuss:
Small example that provides some context:
```golang
               	cfg := &ffresty.Config{}
                secretName := "mtls-config"
                secret := &corev1.Secret{}
		err := ctrlClient.Get(context, types.NamespacedName{Namespace: "default", Name: secretName}, secret)
		if err != nil {
			return nil, err
		}

		tls, err := fftls.NewTLSConfig(context, &fftls.Config{
			Enabled: true,
			CA:      string(secret.Data["ca.crt"]),
			Cert:    string(secret.Data["tls.crt"]),
			Key:     string(secret.Data["tls.key"]),
		}, fftls.ClientType)
		if err != nil {
			return nil, err
		}
		cfg.TLSClientConfig = tls
		// proceed to use ffresty config to make a client
```

pkg/fftls/fftls_test.go
```diff
@@ -24,9 +24,11 @@ import (
 	"crypto/x509"
 	"crypto/x509/pkix"
 	"encoding/pem"
+	"github.com/stretchr/testify/require"
```
nguyer, line 27: Is there a specific reason to pull in `require` here, as opposed to `assert`? I think for unit tests we typically have stuck with `assert`, though I am noticing that is not _as_ true in this particular repo

pkg/fftls/config.go
```diff
@@ -23,14 +23,21 @@ import (
 const (
 	// HTTPConfTLSCAFile the TLS certificate authority file for the HTTP server
 	HTTPConfTLSCAFile = "caFile"
+	// HTTPConfTLSCA the TLS certificate authority in PEM format
+	HTTPConfTLSCA = "ca"
```
nguyer, line 27: I think we should document (probably in this comment so it's visible to the developer) that if these are set they will take precedence over the "file" fields.

nguyer (approved):
Just a couple of minor suggestions but otherwise looks good.

pkg/fftls/fftls_test.go
hfuss: So `require` fails the test immediately whereas `assert` keep runs the test if it doesn't pass. So I always use `require` for anything involve setup, where if it errors / is not successful, the entire things fails.

It's just syntactic sugar for:
```golang
if err != nil {
  t.Fatal(err)
}
```
I believe

pkg/fftls/config.go
hfuss: Comments updated

nguyer: You're right sorry. Had the switch statement backwards in my head.

PR #134, opened 2024-04-05
Do not add alias prefix to filter fields that are already qualified

awrichar:
If a filter field contains "." then assume it is already qualified to the correct scope (similar to how we treat SELECT fields).


peterbroadhurst (approved):
👍

PR #137, opened 2024-05-07
Add custom ID field and AfterLoad hook

awrichar:


pkg/dbsql/crud.go
```diff
@@ -728,6 +740,12 @@ func (c *CrudBase[T]) getManyScoped(ctx context.Context, tableFrom string, fi *f
 		if err != nil {
 			return nil, nil, err
 		}
+		if c.AfterLoad != nil {
+			err = c.AfterLoad(ctx, inst)
```
peterbroadhurst, line 744: Feels still a bit weird that these fields will _not_ get propagated to an object passed into `Insert` or `Update`.

Don't think we can do anything about it, but wonder if any health warnings are needed?

peterbroadhurst (approved):
Approved with one comment to consider @awrichar

pkg/dbsql/crud.go
awrichar: This ends up being most useful for performing final validation and formatting on the whole object after it is loaded (for things that cannot be done in GetFieldPtr). I can add a note to this effect. 

It is less useful for purely generated fields, for the reason you noted. Not sure if I need to add additional comments to specifically discourage that?

awrichar: I guess my unit test is now misleading in this regard.

awrichar: Added a comment and tweaked the unit test

PR #138, opened 2024-05-13
DB-side UPSERT optimiziation support

peterbroadhurst:
... still a little more work to do on this one


pkg/dbsql/postgres_helpers.go
```diff
@@ -0,0 +1,47 @@
...
+	update := sq.Update("REMOVED_BEFORE_RUNNING" /* cheat to avoid table name */)
+	for _, c := range updateCols {
+		update = update.Set(c, values[c])
+	}
+	updateSQL, updateValues, err := update.ToSql()
+	updateSQL, ok := strings.CutPrefix(updateSQL, "UPDATE REMOVED_BEFORE_RUNNING ")
+	if err != nil || !ok {
+		return insert, i18n.NewError(ctx, i18n.MsgDBErrorBuildingStatement, err)
+	}
+	return insert.Suffix(fmt.Sprintf("ON CONFLICT (%s) DO UPDATE", idColumn)).SuffixExpr(sq.Expr(updateSQL, updateValues...)).Suffix("RETURNING " + returnCol), nil
```
awrichar, line 45: I believe this alternate cheat would also work (using a single space as the table name, and then not performing any stripping):

	insert = sq.Insert(table).Columns(insertCols...).Values(insertValues...)
	update := sq.Update(" " /* cheat to avoid table name */)
	for _, c := range updateCols {
		update = update.Set(c, values[c])
	}
	updateSQL, updateValues, err := update.ToSql()
	if err != nil {
		return insert, i18n.NewError(ctx, i18n.MsgDBErrorBuildingStatement, err)
	}
	return insert.Suffix(fmt.Sprintf("ON CONFLICT (%s) DO", idColumn)).
		SuffixExpr(sq.Expr(updateSQL, updateValues...)).
		Suffix("RETURNING " + returnCol), nil

Happy for you to keep it as is or switch it (not sure either is better/worse - just noting it as a possibility).

awrichar (approved):

PR #139, opened 2024-05-14
[log] WithLogFields for Configuring Larger Logging Contexts

onelapahead:
In several case, one might want to set several key value pairs on the context. Currently you have to chain together nested `WithLogField` funcs to do so, which isn't very readable or user friendly.

This allows the users to do so with just a list of strings. Otherwise we could require a `logrus.Fields` struct to be directly provided if we want to avoid the odd-number-check-and-panic ?

### DIFF
--- a/pkg/log/log.go
+++ b/pkg/log/log.go
@@ -48,6 +48,24 @@ func WithLogField(ctx context.Context, key, value string) context.Context {
 	return WithLogger(ctx, loggerFromContext(ctx).WithField(key, value))
 }
 
+func WithLogFields(ctx context.Context, keyValues ...string) context.Context {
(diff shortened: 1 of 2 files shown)


pkg/log/log.go
```diff
@@ -48,6 +48,24 @@ func WithLogField(ctx context.Context, key, value string) context.Context {
 	return WithLogger(ctx, loggerFromContext(ctx).WithField(key, value))
 }
 
+func WithLogFields(ctx context.Context, keyValues ...string) context.Context {
```
EnriqueL8, line 51: Having `...string` is fine to be consistent with `WithLogField` but it doesn't stop us having `logrus.Fields` one

PR #141, opened 2024-05-30
Adding support for leaky bucket throttling (rps + burst) in ffresty client

Chengxuan:
Adding `throttle` section as a configuration so the ffresty client can submit requests at certain rates using leaky bucket style throttling control.

**Note:** this option is not optimal if other clients are using the REST endpoint under the same rate limit bucket.

To handle scenarios like that, we should update ffresty client to be Retry-after header aware or allow users to provide custom handling logic if they use different standards.


pkg/ffresty/config.go
```diff
@@ -67,6 +67,8 @@ const (
 	HTTPIdleTimeout = "idleTimeout"
 	// HTTPMaxIdleConns the max number of idle connections to hold pooled
 	HTTPMaxIdleConns = "maxIdleConns"
+	// HTTPRPS the max number of request to submit per second, default value is 0, which turns off the RPS control
+	HTTPRPS = "rps"
```
EnriqueL8, line 71: Would prefer more explicit configuration name
```suggestion
	HTTPRPS = "maxRequestsPerSecond"
```

pkg/ffresty/ffresty.go
```diff
@@ -148,6 +150,12 @@ func NewWithConfig(ctx context.Context, ffrestyConfig Config) (client *resty.Cli
 		client = resty.NewWithClient(httpClient)
 	}
 
+	var rpsLimiter *rate.Limiter
```
EnriqueL8, line 153: There is a native rate limiter that can be associated with the resty client https://github.com/go-resty/resty/pull/715/files so we wouldn''t need to call Wait ourselves. I think this is neater

EnriqueL8: Would look like 
```
client.SetRateLimiter(rate.NewLimiter(rate.Limit(rps), rps)
```

EnriqueL8 (changes requested):

pkg/ffresty/ffresty.go
Chengxuan: the resty rate limit fails the request if the rate exceeded the limit while the behaviour introduced in this PR use the native golang rate limiter to wait.

pkg/ffresty/config.go
Chengxuan: Is it because it's too short, or developers using ffresty wouldn't understand what RPS is (like they would do for URL/TLS)?  `rps` is a common attribute for most ingress configurations, that's the reason i keep it consistent.

Chengxuan:
@EnriqueL8 thanks for the review. Replied to your comments

pkg/ffresty/ffresty.go
EnriqueL8: ah I see - so it's like a buffer, we wait a max of a second until the next requests go through

EnriqueL8: Chatted that the context used should be the request context

pkg/ffresty/ffresty.go
```diff
@@ -210,6 +213,10 @@ func NewWithConfig(ctx context.Context, ffrestyConfig Config) (client *resty.Cli
 		client = resty.NewWithClient(httpClient)
 	}
 
+	if ffrestyConfig.RequestPerSecond != 0 { // NOTE: 0 is treated as RPS control disabled
+		rateLimiter = rate.NewLimiter(rate.Limit(ffrestyConfig.RequestPerSecond), ffrestyConfig.RequestPerSecond)
```
peterbroadhurst, line 217: I think related to above comment `r` and `b` here should be separate configurations, with separate defaults

peterbroadhurst (changes requested):
Really great that there's a leaky bucket algorithm built into Go to plug in here 🎉 

One request on the configuration split.

Chengxuan:
@peterbroadhurst comments addressed

peterbroadhurst (approved):
Thanks @Chengxuan for the extra spelling round there - looks great

Issue #142, opened 2024-05-30
[feature] Resty Metrics for HTTP vs TCP Failures


onelapahead:
With this library providing a metrics manager for Prometheus, and autoconfiguration for Resty clients, we should support automatically wiring up metrics from Resty clients into Prometheus.

Setting it up is a bit tricky, as the Resty client metrics must be registered only once per Prometheus registry, which the manager should handle. Clients should all refer to same global Resty client metrics / subsystem. When creating a Resty client it should be optional (possibly a separate `New` func) to provide a metrics manager and registry that it then wires itself into the client, and ensures that the metrics it depends on are registered or will be registered.

We can then use Resty's `OnError` (https://github.com/go-resty/resty?tab=readme-ov-file#onerror-hooks) to receive a hook when errors occur, determine if the error is a response error or otherwise assume it was a network error, and increment an `http_failure` vs `network_failure` counter depending. These metrics do not need specific URLs yet as that would be high cardinality.

If we wish to distinguish network errors with specific known errors and tags we could do something like: https://stackoverflow.com/questions/67646842/how-to-handle-net-errors-in-golang-properly

Once this metrics framework and basic failure counters are in place, we can consider expanding support for response time histograms, counters for particular HTTP methods and response codes, etc. leveraging `OnBeforeRequest` and `OnAfterResponse` more.

alexwood:
Merging the above proposals I am interpretting the proposal as ...

1. enhance ffresty to have a (global) EnableClientMetrics func which configures metrics manager and registry globally for use in FFResty clients.
2. enhance ffresty to support global hooks for onError. 
3. enhance ffresty to add specific global hooks for metrics if they are configured via (1) using the global onError hooks and possibly other global hooks (2)

onelapahead:
That proposal sounds great to me @alexwood thanks! Can iterate through the details once a draft PR is ready and we have a working example of the failure metrics as prove point for hooks / metrics wiring framework 🙇🏻

alexwood:
Changes merged that add http response metrics and also non http error metrics

@onelapahead can we close this issue or do you wish to keep it open for some additional follow on requirements we want to include.

EnriqueL8:
I think we can close this one! Thanks for the contribution @alexwood and @onelapahead

PR #143, opened 2024-06-04
ffresty metrics and hooks

alexwood:
1.  register static onError/onSuccess hooks using RegisterGlobalOnError/RegisterGlobalOnSuccess applies to all clients in an application
2. enable http metrics  EnableClientMetrics applies to all clients in an application - currently adds onError and onSuccess hooks and emits a count of http response with status code and for non http errors a separate count

for https://github.com/hyperledger/firefly-common/issues/142

### DIFF
--- a/pkg/ffresty/ffresty.go
+++ b/pkg/ffresty/ffresty.go
@@ -94,7 +94,7 @@ func EnableClientMetrics(ctx context.Context, metricsRegistry metric.MetricsRegi
 			return err
 		}
 		metricsManager.NewCounterMetricWithLabels(ctx, "http_response", "HTTP response", []string{"status", "error", "host", "method"}, false)
-		metricsManager.NewCounterMetricWithLabels(ctx, "network_error", "Network error", []string{"host", "method"}, false)	
+		metricsManager.NewCounterMetricWithLabels(ctx, "network_error", "Network error", []string{"host", "method"}, false)
 	}
 
 	// create hooks
--- a/pkg/ffresty/ffresty.go
+++ b/pkg/ffresty/ffresty.go
@@ -24,6 +24,7 @@ import (
 	"io"
 	"net"
 	"net/http"
+	"net/url"
 	"regexp"
 	"strings"
 	"time"
@@ -34,6 +35,7 @@ import (
 	"github.com/hyperledger/firefly-common/pkg/fftypes"
 	"github.com/hyperledger/firefly-common/pkg/i18n"
 	"github.com/hyperledger/firefly-common/pkg/log"
+	"github.com/hyperledger/firefly-common/pkg/metric"
 	"github.com/sirupsen/logrus"
 )
 
@@ -49,6 +51,12 @@ type Config struct {
 	URL string `json:"httpURL,omitempty"`
 	HTTPConfig
 }
+
+var (
+	metricsManager metric.MetricsManager
+	onErrorHooks   []resty.ErrorHook
+	onSuccessHooks []resty.SuccessHook
+)
 
 // HTTPConfig is all the optional configuration separate to the URL you wish to invoke.
 // This is JSON serializable with docs, so you can embed it into API objects.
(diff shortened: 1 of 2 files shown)


pkg/ffresty/ffresty.go
```diff
@@ -50,6 +51,13 @@ type Config struct {
 	HTTPConfig
 }
 
+var (
+	metricsManager metric.MetricsManager
+	onErrorHooks []func(*resty.Request, error)
```
onelapahead, line 56: This looks great to me, think we want to make some comments indicating we do not protect these lists with mutex's bc we only expect `Register` funcs to be called on startup.

nit: lets make a types for these funcs? i.e.`type ErrorProcessor func(*resty.Request, error)` ?

pkg/ffresty/ffresty.go
```diff
@@ -77,6 +85,43 @@ type HTTPConfig struct {
 	OnBeforeRequest               func(req *resty.Request) error            `json:"-"` // called before each request, even retry
 }
 
+func EnableClientMetrics(ctx context.Context, metricsRegistry metric.MetricsRegistry) error {
+	//create a metrics manager
+	mm, err := metricsRegistry.NewMetricsManagerForSubsystem(ctx, "ffresty")
```
onelapahead, line 90: I think by default all metrics made through our lib get prefixed with `ff` so maybe this subsystem can just be called `resty` ?

Otherwise think its fine that the resty clients have a dedicated manager, and then users just need to provide a registry which is presumably wired up to their metrics server to expose the client metrics.

pkg/ffresty/ffresty.go
```diff
@@ -77,6 +85,43 @@ type HTTPConfig struct {
...
+	if err != nil {
+		return err
+	}
+
+	metricsManager := mm
+	metricsManager.NewCounterMetricWithLabels(ctx, "http_response", "HTTP response", []string{"status","error"}, false)
```
onelapahead, line 96: I know I said in the issue that a `url` tag might be too high in cardinality, but what about a `host` tag ?

That should be lower in cardinality since it won't include all the paths permutations, but very helpful for determining if HTTP / TCP errors are for particular hosts / backends or not.

onelapahead (commented):
Overall approved, just a few small points. Ty for the great unit test!

pkg/ffresty/ffresty.go
alexwood: used the built in one from resty itself

pkg/ffresty/ffresty.go
alexwood: changed removed the ff

alexwood:
So playing with dashboards and not seeing the status code for errors :( they come out with status code `0`  not sure why will have to add some trace logging to figure it out.

I think what I really want is to see the 500/502/503 etc coming out here.

alexwood:
More improvements as discussed - tested in an apigw application that calls multiple hosts  and can see useful metrics for both response codes and response time coming out

Screenshot 2024-06-05 at 09 43 44

pkg/ffresty/ffresty.go
```diff
@@ -50,6 +52,13 @@ type Config struct {
 	HTTPConfig
 }
 
+
+var (
+	metricsManager metric.MetricsManager
+	onErrorHooks   []resty.ErrorHook
+	onSuccessHooks []resty.SuccessHook
+)
```
EnriqueL8, line 59: This feels like an anti pattern in Golang storing into vars as this will be for the whole package? Shouldn't these be nested under the client?

EnriqueL8: And I think it should be driven through the config to instantiate one per client?

pkg/ffresty/ffresty.go
```diff
@@ -204,9 +270,11 @@ func NewWithConfig(ctx context.Context, ffrestyConfig Config) (client *resty.Cli
 	})
 
 	// Note that callers using SetNotParseResponse will need to invoke this themselves
-
 	client.OnAfterResponse(func(c *resty.Client, r *resty.Response) error { OnAfterResponse(c, r); return nil })
 
+	client.OnError( func(req *resty.Request, e error) { OnError(req, e) })
+	client.OnSuccess(func(c *resty.Client, r *resty.Response) { OnSuccess(c, r) })
```
EnriqueL8, line 275: I wonder if this should be a flag to enable metrics on the config as users might override this afterwards if they wish?

EnriqueL8 (commented):
Mainly around if this should be config option and not having a "global" enable metrics manager. I get that you need to pass in the manager to it and we return a resty.Client not a ffresty interface that we could extend to enable metrics


Update missed the issue comment, it sounds like we want to have it global for adding hooks https://github.com/hyperledger/firefly-common/issues/142#issuecomment-2140071723 and since we do not have a FFResty object we need to use these package vars to store them

pkg/ffresty/ffresty.go
EnriqueL8: This needs to be always be here by default to enable adding hooks that apply to previously created clients, so all good

pkg/ffresty/ffresty.go
alexwood: this was part of the intention of this change that this is globally statically configured not per instance

alexwood: see https://github.com/hyperledger/firefly-common/issues/142 for explanation

pkg/ffresty/ffresty.go
```diff
@@ -77,6 +85,51 @@ type HTTPConfig struct {
...
 }
 
+func EnableClientMetrics(ctx context.Context, metricsRegistry metric.MetricsRegistry) error {
+	// create a metrics manager
+	mm, err := metricsRegistry.NewMetricsManagerForSubsystem(ctx, "resty")
+	if err != nil {
+		return err
+	}
+	metricsManager = mm
+	metricsManager.NewGaugeMetricWithLabels(ctx, "http_duration", "HTTP duration", []string{"status", "error", "host", "method"}, false)
```
onelapahead, line 95: So `duration` would be best as a Histogram with a few buckets as opposed to a guage.

From there, we can calculate average response time and estimate percentiles if we want but it is high cardinality, and the buckets is likely something we'd want to expose as config on this method or globally.

alexwood: removed the gauge metric

onelapahead (commented):
looks good! ty

EnriqueL8 (approved):
LGTM

pkg/ffresty/ffresty.go
```diff
@@ -77,6 +85,50 @@ type HTTPConfig struct {
 	OnBeforeRequest               func(req *resty.Request) error            `json:"-"` // called before each request, even retry
 }
 
+func EnableClientMetrics(ctx context.Context, metricsRegistry metric.MetricsRegistry) error {
+	// create a metrics manager
+	mm, err := metricsRegistry.NewMetricsManagerForSubsystem(ctx, "resty")
+	if err != nil {
+		return err
```
peterbroadhurst, line 92: UT test gap on this line

pkg/ffresty/ffresty.go
```diff
@@ -77,6 +85,50 @@ type HTTPConfig struct {
...
+
+	// create hooks
+	onErrorMetricsHook := func(req *resty.Request, err error) {
+		method := req.Method
+		u, _ := url.Parse(req.URL)
+		host := u.Host
+		if v, ok := err.(*resty.ResponseError); ok {
+			if v.Response != nil && v.Response.RawResponse != nil {
+				code := v.Response.RawResponse.StatusCode
+				metricsManager.IncCounterMetricWithLabels(ctx, "http_response", map[string]string{"status": fmt.Sprintf("%d", code), "error": "true", "host": host, "method": method}, nil)
```
peterbroadhurst, line 106: UT test gap on this line

peterbroadhurst (changes requested):
This contribution looks great @alexwood 
I like that you've taken the `RegisterGlobal...` approach, but also provided a pre-baked example for metrics that can be used by a consumer of this library that is buying into the metrics stack that's also in `firefly-common`.

That aligns really well with the spirit of the utilities in this package.

The only request before merge from me is that to keep the coverage up on the repo, as we've found it helpful to keep a 100% coverage on these utilities.

pkg/ffresty/ffresty.go
alexwood: additional tests added

pkg/ffresty/ffresty.go
alexwood: so this proved impossible to mock since net/http deliberately nils the response if the error is set. 
reading the net/http code their was a fringe case where it did allow it through and I was able to hit this in full stack testing, but in such case the statusCode was `0` so not useful
Therefore I decided for OnError to now log http_response metric and only log the network_error metric.

alexwood: updated now coverage should be 100%

alexwood:
@peterbroadhurst  this is ready for re-review - thanks

peterbroadhurst (approved):

PR #145, opened 2024-08-16
fix: change SwaggerGen to support multiple values with the same path

dwertent:
This PR offers an alternative solution to the issue addressed in [this PR](https://github.com/hyperledger/firefly/pull/1558). Rather than modifying the API, we've adjusted the behavior of the Swagger generation to ensure the value is no longer ignored.


pkg/ffapi/openapi3.go
```diff
@@ -389,12 +389,15 @@ func (sg *SwaggerGen) addFilters(ctx context.Context, route *Route, op *openapi3
 func (sg *SwaggerGen) addRoute(ctx context.Context, doc *openapi3.T, route *Route) {
 	var routeDescription string
 	pi := sg.getPathItem(doc, route.Path)
+	if route.Path == "" {
+		fmt.Println("route.Path is empty")
+	}
```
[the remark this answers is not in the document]
dwertent: This was pushed by mistake :)

pkg/ffapi/openapi3.go
```diff
@@ -116,7 +116,7 @@ func (sg *SwaggerGen) getPathItem(doc *openapi3.T, path string) *openapi3.PathIt
 	if doc.Paths == nil {
 		doc.Paths = &openapi3.Paths{}
 	}
-	pi := doc.Paths.Find(path)
+	pi := doc.Paths.Value(path)
```
EnriqueL8, line 119: Just being careful with this, as it would affect all the swagger documents generated from now on

Is there a scenario where normalising does make sense? I tried to find the history in the `kin-openapi` repo and it seems that it was there from the original commit

EnriqueL8 (commented):
Just want to understand where normalising the path makes sense

pkg/ffapi/openapi3.go
dwertent: Ok, we can wait for a response [here](https://github.com/OAI/OpenAPI-Specification/discussions/4036).

EnriqueL8 (approved):
Looks good - we need to be careful when pulling this into other repos and make sure the changes are correct

PR #148, opened 2024-09-06
feat: Add DeepCopy method to JSONObject

dwertent:
The `DeepCopy` method is added to the `JSONObject` struct in order to create a deep copy of the object. This method recursively copies the map and slice elements of the object to ensure that the copied object is independent of the original. This is useful after encountering [this issue](https://github.com/hyperledger/firefly/issues/1310) that occurred because of a shallow copy.

[This code](https://github.com/hyperledger/firefly/pull/1561/files#diff-4e70312913480467cb8f47462da4c1bc568a2d750f22519a1bae78ad41a4b197R106) should call `cop.Input = op.Input.DeepCopy()` once this code is published.


EnriqueL8 (approved):
Looks good - not sure why we didn't merge this before!

PR #151, opened 2024-09-13
Route handler to use UseNumber decoding

matthew1001:
Relates to https://github.com/hyperledger/firefly/issues/1568 and follows on from https://github.com/hyperledger/firefly-common/pull/147.

Like the changes in that PR, this code path also needs to parse large numbers into `json.Number` instead of `float64`.


EnriqueL8 (approved):
Great catch! Looks good to me

PR #152, opened 2024-10-14
Provide the client certificate without relying on golang matching it

EnriqueL8:
Noticed when providing the client certificates to ffresty, it wasn't getting being passed on the API call itself and it's due to Golang not matching that cert based on the requirements of the request. 

Some background https://forum.golangbridge.org/t/issue-with-client-tls-authentication/8214


peterbroadhurst (approved):
👍 - a fun "feature" of the Go TLS implementation

PR #153, opened 2024-10-22
[ffresty] [bug] Set maxConnsPerHost Based on Config

onelapahead:
The default is 0, and since this was empty / not set it was still defaulting to 0 anyways. But this ensures if a user does provide a non-zero maxConnPerHost, it will be respected.


peterbroadhurst (approved):
👍

PR #154, opened 2024-10-22
[ffresty] maxIdleConnsPerHost Setting

onelapahead:
Follow's up #153. Reading common performance problems in Go `net/http`, the `maxConnsPerHost` is helpful being unlimited by default but it sounds like the `maxIdleConnsPerHost` will always default to 2 based on the source:
```golang
// DefaultMaxIdleConnsPerHost is the default value of [Transport]'s
// MaxIdleConnsPerHost.
const DefaultMaxIdleConnsPerHost = 2
// ...
type Transport struct {
  // ...
  // MaxIdleConnsPerHost, if non-zero, controls the maximum idle
  // (keep-alive) connections to keep per-host. If zero,
  // DefaultMaxIdleConnsPerHost is used.
  MaxIdleConnsPerHost int
  // ...
}
```

Based on https://stackoverflow.com/a/31810116 and https://www.reddit.com/r/golang/comments/1730po3/comment/k51x8jn/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button we should see our clients default to creating many more connections per host under higher load.

### DIFF
--- a/pkg/ffresty/config.go
+++ b/pkg/ffresty/config.go
@@ -32,7 +32,8 @@ const (
 	defaultRequestTimeout                = "30s"
 	defaultHTTPIdleTimeout               = "475ms" // Node.js default keepAliveTimeout is 5 seconds, so we have to set a base below this
 	defaultHTTPMaxIdleConns              = 100     // match Go's default
-	defaultHTTPMaxConnsPerHost           = 0
+	defaultHTTPMaxConnsPerHost           = 0       // unlimited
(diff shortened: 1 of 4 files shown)


EnriqueL8 (approved):
Looks good - thanks for the clear explanation and links

peterbroadhurst (approved):
Understand and agree with this change.

Including the fact it changes default behavior for all consumers of this library.

PR #155, opened 2024-10-24
[pkg/metric] - Add the ability to register a new metrics collector to the registry

calbritt:
Adding RegisterCollector to support adding a custom prometheus Collector to the prometheus Metrics Registry. Currently `MetricsRegistry` only allows creating a registry with the default GoCollector and ProcessCollector.

### DIFF
--- a/pkg/metric/metric.go
+++ b/pkg/metric/metric.go
@@ -59,7 +59,7 @@ type MetricsRegistry interface {
 	// GetHTTPMetricsInstrumentationsMiddlewareForSubsystem returns the HTTP middleware of a subsystem that used predefined HTTP metrics
 	GetHTTPMetricsInstrumentationsMiddlewareForSubsystem(ctx context.Context, subsystem string) (func(next http.Handler) http.Handler, error)
 
-	RegisterCollector(ctx context.Context, collectors prometheus.Collector)
+	RegisterCollector(ctx context.Context, collector prometheus.Collector)
 }
 
 type FireflyDefaultLabels struct {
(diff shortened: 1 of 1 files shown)


pkg/metric/metric.go
```diff
@@ -58,6 +58,8 @@ type MetricsRegistry interface {
 	NewHTTPMetricsInstrumentationsForSubsystem(ctx context.Context, subsystem string, useRouteTemplate bool, reqDurationBuckets []float64, labels map[string]string) error
 	// GetHTTPMetricsInstrumentationsMiddlewareForSubsystem returns the HTTP middleware of a subsystem that used predefined HTTP metrics
 	GetHTTPMetricsInstrumentationsMiddlewareForSubsystem(ctx context.Context, subsystem string) (func(next http.Handler) http.Handler, error)
+
+	RegisterCollector(ctx context.Context, collectors prometheus.Collector)
```
onelapahead, line 62: nit: collector

pkg/metric/metric.go
```diff
@@ -185,3 +187,7 @@ func (pmr *prometheusMetricsRegistry) GetHTTPMetricsInstrumentationsMiddlewareFo
 	}
 	return httpInstrumentation.Middleware, nil
 }
+
+func (pmr *prometheusMetricsRegistry) RegisterCollector(ctx context.Context, collectors prometheus.Collector) {
```
onelapahead, line 191: nit: let's name this `MustRegisterCollector`. The `Must` is a Go-ism to indicate - this will panic rather than return an error if it fails.

Arguably we should have a `RegisterCollector` that calls the non-must version of the func and bubbles up the error.

onelapahead (commented):
Looks great to me, just some small nits. Will defer to @EnriqueL8 @shorsher  for a review.

pkg/metric/metric.go
calbritt: Yes, I had wondered the same on using `Register` instead but followed the MustRegister pattern with the default collectors on the registry - https://github.com/hyperledger/firefly-common/pull/155/files#diff-8b48568f0ecef3176c9d0ccd9ace7f9c73753c2a0ad6fd03f99c8f93ac186d73R98

peterbroadhurst (approved):

PR #157, opened 2024-12-05
[metrics] Configurable Prefix and Avoiding 'ff_' in Tags

onelapahead:
## Changes

Makes it so the `ff_` prefix on metrics can be replaced with a prefix of the user's choosing via a new optional `Options` struct. And then drops the `ff_` prefix on the metrics label `component` so that it is consistent with other labeling schemes from logs or other metrics libraries, and saves some bytes per metric. Made it so theres a new `New` func for backwards compatibility.

However, having this metric may be redundant bc it can often be inferred from a target's hostname or pod labels. So we should consider dropping, or making the `component` label optional.

### DIFF
--- a/pkg/ffresty/metrics.go
+++ b/pkg/ffresty/metrics.go
@@ -1,0 +1,1 @@
+package ffresty
--- a/pkg/ffresty/metrics.go
+++ b/pkg/ffresty/metrics.go
@@ -1,1 +1,0 @@
-package ffresty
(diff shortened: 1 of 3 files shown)


pkg/ffresty/metrics.go
```diff
@@ -0,0 +1 @@
+package ffresty
```
EnriqueL8, line 1: Did you mean to commit this? This is just missing copyright and the CI is complaining

EnriqueL8 (commented):
Looks good - just one comment

pkg/ffresty/metrics.go
onelapahead: Yeah will remove

EnriqueL8 (commented):
Looks good - probably worth updating the tests

PR #158, opened 2024-12-09
Implement throttling for WS client

Chengxuan:
The configurations already exist due to reusing ffresty config initialisation logic.

This PR implements the throttling logic itself

### DIFF
--- a/pkg/ffresty/ffresty.go
+++ b/pkg/ffresty/ffresty.go
@@ -54,7 +54,7 @@ type Config struct {
 }
 
 var (
-	rateLimiter    *rate.Limiter
+	rateLimiterMap map[*resty.Client]*rate.Limiter
 	metricsManager metric.MetricsManager
 	onErrorHooks   []resty.ErrorHook
 	onSuccessHooks []resty.SuccessHook
@@ -177,7 +177,7 @@ func New(ctx context.Context, staticConfig config.Section) (client *resty.Client
 	return NewWithConfig(ctx, *ffrestyConfig), nil
 }
 
-func getRateLimiter(rps, burst int) *rate.Limiter {
+func GetRateLimiter(rps, burst int) *rate.Limiter {
 	if rps != 0 { // if rps is not set no need for a rate limiter
 		rpsLimiter := rate.Limit(rps)
 		if burst == 0 {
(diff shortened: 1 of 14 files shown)


pkg/fftypes/uuid.go
```diff
@@ -93,11 +93,13 @@ func (u *UUID) UnmarshalBinary(b []byte) error {
 }
 
 func (u *UUID) HashBucket(buckets int) int {
-	if u == nil {
+	if u == nil || buckets <= 0 {
```
EnriqueL8, line 96: why? the next line would NPE if buckets > 0

pkg/ffresty/ffresty.go
```diff
@@ -54,7 +54,7 @@ type Config struct {
 }
 
 var (
-	rateLimiter    *rate.Limiter
+	rateLimiterMap map[*resty.Client]*rate.Limiter
```
EnriqueL8, line 57: confused about this one.. we are using a local variable so where we overriding this rate limiter each time we created a new client and you are fixing this bug?

pkg/ffresty/ffresty.go
```diff
@@ -242,9 +245,9 @@ func NewWithConfig(ctx context.Context, ffrestyConfig Config) (client *resty.Cli
 	client.SetTimeout(time.Duration(ffrestyConfig.HTTPRequestTimeout))
 
 	client.OnBeforeRequest(func(_ *resty.Client, req *resty.Request) error {
-		if rateLimiter != nil {
+		if rateLimiterMap[client] != nil {
```
EnriqueL8, line 248: I wonder if the rate limiter can be a property of the resty client ... seems like it is there https://github.com/go-resty/resty/pull/715/files

EnriqueL8: ah the way is implemented is different, we wait and the implementation of resty throws an error

EnriqueL8 (changes requested):
Looking good - a few questions

pkg/ffresty/ffresty.go
Chengxuan: yes this is a bug fix

pkg/ffresty/ffresty.go
Chengxuan: i think this is resolved

pkg/fftypes/uuid.go
Chengxuan: converting a negative integer to uint doesn't make sense.

EnriqueL8: ah sorry yeah makes sense

EnriqueL8 (approved):
Looks good

PR #160, opened 2024-12-12
[ffresty] [metric] HTTP Response Time and Complete Gauge Support

onelapahead:
For Resty, we enhance the existing before/after hooks so that we record the elapsed time on responses. And use more labels and smarter context's for consistently deriving any request metadata like the host.

Then, within the metric manager itself, we didn't offer the ability to use gauge inc/dec which can be very useful rather than using set.


pkg/ffresty/ffresty.go
```diff
@@ -40,7 +40,15 @@ import (
 	"golang.org/x/time/rate"
 )
 
+const (
+	metricsHTTPResponsesTotal  = "http_responses_total"
+	metricsNetworkErrorsTotal  = "network_errors_total"
+	metricsHTTPRequestBodySize = "http_request_body_size_bytes"
```
onelapahead, line 46: rm

firefly-common.iml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
onelapahead, line 1: rm

pkg/ffresty/ffresty.go
```diff
@@ -40,7 +40,14 @@ import (
 	"golang.org/x/time/rate"
 )
 
+const (
+	metricsHTTPResponsesTotal = "http_responses_total"
+	metricsNetworkErrorsTotal = "network_errors_total"
+	metricsHTTPResponseTime   = "http_response_time_seconds"
```
Chengxuan, line 46: Some size metrics for data transmitted would be useful to add as well.

Chengxuan (commented):
That is a great step forward, thanks @onelapahead .

I asked a question about the metrics type for the response time. Switching to histogram will have the following benefits, which could be compelling:
1. reduce observation costs on the client
2. enable aggregation across instances

^^ from https://prometheus.io/docs/practices/histograms/

pkg/ffresty/ffresty.go
```diff
@@ -98,27 +105,21 @@ func EnableClientMetrics(ctx context.Context, metricsRegistry metric.MetricsRegi
 		if err != nil {
 			return err
 		}
-		metricsManager.NewCounterMetricWithLabels(ctx, "http_response", "HTTP response", []string{"status", "error", "host", "method"}, false)
-		metricsManager.NewCounterMetricWithLabels(ctx, "network_error", "Network error", []string{"host", "method"}, false)
+		metricsManager.NewCounterMetricWithLabels(ctx, metricsHTTPResponsesTotal, "HTTP response", []string{"status", "error", "host", "method"}, false)
+		metricsManager.NewCounterMetricWithLabels(ctx, metricsNetworkErrorsTotal, "Network error", []string{"host", "method"}, false)
+		metricsManager.NewSummaryMetricWithLabels(ctx, metricsHTTPResponseTime, "HTTP response time", []string{"status", "host", "method"}, false)
```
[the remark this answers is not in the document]
onelapahead: Histogram is more costly in terms of cardinality bc of all the buckets you make. So this felt light enough to only calculate _average_ response time across a few basic dimensions.

Would love to toggle btwn histogram and summaries based on certain settings eventually, but that was a bigger change than I had an appetite for.

onelapahead:
> enable aggregation across instances

To clarify - this is still entirely possible with summaries. They are similar to a histogram but less counters:

```
# summary example
http_response_time_sum{}
http_response_time_count{}

# histogram example
http_response_time_buckets{le="1.0"}
http_response_time_buckets{le="0.5"}
http_response_time_buckets{le="0.1"}
http_response_time_sum{}
http_response_time_count{}
```

Histograms are great/preferred for aggregate percentiles and therefore performance engineering. Summaries are great for aggregated averages and therefore reliability engineering.

_Quantiles_ are a special version of summary, which to your point are basically worthless at any real scale bc they cannot be aggregated. However, they are less expensive cardinality than histograms. So if you care about individual "thing's" performance they are useful.

Do hope ff-common can offer histograms as an option, but not by default since they are quite expensive. But will save that for a future contributor.

Chengxuan (approved):

EnriqueL8 (commented):
This looks good @onelapahead 

Getting `Error: pkg/ffresty/ffresty.go:201:1: cyclomatic complexity 37 of func `NewWithConfig` is high (> 30) (gocyclo)` from build, given this function is doing a lot of just configuration we can probably ignore this complexity warning by adding the ignore comment. Or if you are up for it we can break it down.

pkg/ffresty/ffresty.go
```diff
@@ -142,13 +143,17 @@ func OnAfterResponse(c *resty.Client, resp *resty.Response) {
...
 	status := resp.StatusCode()
 	if status >= 300 {
 		level = logrus.ErrorLevel
 	}
-	log.L(rCtx).Logf(level, "<== %s %s [%d] (%.2fms)", resp.Request.Method, resp.Request.URL, status, elapsed)
+	log.L(rCtx).Logf(level, "<== %s %s [%d] (%dms)", resp.Request.Method, resp.Request.URL, status, time.Since(rc.start).Milliseconds())
+	if metricsManager != nil {
+		metricsManager.ObserveSummaryMetricWithLabels(rCtx, metricsHTTPResponseTime, elapsed.Seconds(), map[string]string{"status": fmt.Sprintf("%d", status), "host": rCtx.Value(hostCtxKey{}).(string), "method": resp.Request.Method}, nil)
+	}
+	// TODO use req.TraceInfo() for richer metrics at the DNS and transport layer
```
EnriqueL8, line 156: leaving this TODO here?

EnriqueL8, line 156: leaving this todo here?

pkg/ffresty/ffresty.go
```diff
@@ -253,6 +258,26 @@ func NewWithConfig(ctx context.Context, ffrestyConfig Config) (client *resty.Cli
...
 		}
 		rCtx := req.Context()
+		// Record host in context to avoid redundant parses in hooks
+		var u *url.URL
+		if req.URL != "" {
+			u, _ = url.Parse(req.URL)
+		}
+		// The req.URL might have only set a path i.e. /home, fallbacking to the base URL of the client.
+		// So if the URL is nil, that's likely the case and we'll derive the host from the configured
+		// based instead.
```
EnriqueL8, line 268:
```suggestion
		// base instead.
```

pkg/ffresty/ffresty.go
Chengxuan: @EnriqueL8  this is a suggestion, it should not block the merge of this PR.  thanks for checking.

EnriqueL8 (commented):
Looking good! Just spotted a few minor things

pkg/ffresty/ffresty.go
onelapahead: correct, dont think its worth of an issue (unless you feel otherwise) but a good reminder for future contributors reading code that more is possible with the metrics and where to get started with it

EnriqueL8: sounds good

EnriqueL8 (approved):
Ready to go! Thanks again @onelapahead for the contribution!

PR #169, opened 2025-02-19
default monitoring routes

Brindrajsinh-Chauhan:
As a part of the new monitoring server, this PR adds some out of the box routes to this server that can be directly used as liveness and readiness probes for various Kubernetes deployments. Similar to some out of the box routes like `/api` that are provided by the apiServer, the new monitoring server provides a default route `/health`. It is configurable similar to metrics path thus allowing uses to define the path if they wish. This can reduce the duplication effort to add special monitoring routes just for such kubernetes probes. User can still add additional monitoring routes they might want to use through the `MonitoringRoutes` config of the API server.

### DIFF
--- a/pkg/ffapi/apiserver.go
+++ b/pkg/ffapi/apiserver.go
@@ -63,6 +63,7 @@ type apiServer[T any] struct {
 	handleYAML                bool
 	monitoringEnabled         bool
 	metricsPath               string
+	monitoringPath            string
 	monitoringPublicURL       string
 	mux                       *mux.Router
 
@@ -102,6 +103,7 @@ func NewAPIServer[T any](ctx context.Context, options APIServerOptions[T]) APISe
 		requestMaxTimeout:         options.APIConfig.GetDuration(ConfAPIRequestMaxTimeout),
 		monitoringEnabled:         options.MonitoringConfig.GetBool(ConfMonitoringServerEnabled),
 		metricsPath:               options.MonitoringConfig.GetString(ConfMonitoringServerMetricsPath),
+		monitoringPath:            options.MonitoringConfig.GetString(ConfMonitoringServerMonitoringPath),
 		alwaysPaginate:            options.APIConfig.GetBool(ConfAPIAlwaysPaginate),
 		handleYAML:                options.HandleYAML,
 		apiDynamicPublicURLHeader: options.APIConfig.GetString(ConfAPIDynamicPublicURLHeader),
@@ -294,6 +296,11 @@ func (as *apiServer[T]) notFoundHandler(res http.ResponseWriter, req *http.Reque
 	return 404, i18n.NewError(req.Context(), i18n.Msg404NotFound)
 }
 
+func (as *apiServer[T]) emptyJSONHandler(res http.ResponseWriter, _ *http.Request) (status int, err error) {
+	res.Header().Add("Content-Type", "application/json")
+	return 200, nil
+}
+
 func (as *apiServer[T]) createMonitoringMuxRouter(ctx context.Context) *mux.Router {
 	r := mux.NewRouter().UseEncodedPath()
 	hf := as.handlerFactory() // TODO separate factory for monitoring ??
@@ -305,9 +312,10 @@ func (as *apiServer[T]) createMonitoringMuxRouter(ctx context.Context) *mux.Rout
 	r.Path(as.metricsPath).Handler(h)
 
 	for _, route := range as.MonitoringRoutes {
-		r.HandleFunc(route.Path, as.routeHandler(hf, route)).Methods(route.Method)
+		r.HandleFunc(fmt.Sprintf("/%s", route.Path), as.routeHandler(hf, route)).Methods(route.Method)
 	}
 
+	r.HandleFunc(as.monitoringPath, hf.APIWrapper(as.emptyJSONHandler))
 	r.NotFoundHandler = hf.APIWrapper(as.notFoundHandler)
 	return r
 }
--- a/pkg/ffapi/apiserver_config.go
+++ b/pkg/ffapi/apiserver_config.go
@@ -22,8 +22,9 @@ import (
 )
 
 var (
-	ConfMonitoringServerEnabled     = "enabled"
-	ConfMonitoringServerMetricsPath = "metricsPath"
+	ConfMonitoringServerEnabled        = "enabled"
+	ConfMonitoringServerMetricsPath    = "metricsPath"
+	ConfMonitoringServerMonitoringPath = "monitoringPath"
 
 	ConfAPIDefaultFilterLimit     = "defaultFilterLimit"
 	ConfAPIMaxFilterLimit         = "maxFilterLimit"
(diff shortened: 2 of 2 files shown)


pkg/ffapi/apiserver.go
```diff
@@ -305,9 +312,10 @@ func (as *apiServer[T]) createMonitoringMuxRouter(ctx context.Context) *mux.Rout
 	r.Path(as.metricsPath).Handler(h)
 
 	for _, route := range as.MonitoringRoutes {
-		r.HandleFunc(route.Path, as.routeHandler(hf, route)).Methods(route.Method)
+		r.HandleFunc(fmt.Sprintf("/%s", route.Path), as.routeHandler(hf, route)).Methods(route.Method)
 	}
 
+	r.HandleFunc(as.monitoringPath, hf.APIWrapper(as.emptyJSONHandler))
```
Chengxuan, line 318: personally, I think `/health` should just be the path, at most a toggle to add/remove that path.

And we move this line above the for loop, so if consumers want to reuse `/health` endpoint, they can override it.

pkg/ffapi/apiserver_config.go
```diff
@@ -49,4 +50,5 @@ func InitAPIServerConfig(apiConfig, monitoringConfig, corsConfig config.Section)
 	httpserver.InitHTTPConfig(monitoringConfig, 6000)
 	monitoringConfig.AddKnownKey(ConfMonitoringServerEnabled, true)
 	monitoringConfig.AddKnownKey(ConfMonitoringServerMetricsPath, "/metrics")
+	monitoringConfig.AddKnownKey(ConfMonitoringServerMonitoringPath, "/health")
```
Chengxuan, line 53:
```suggestion
	monitoringConfig.AddKnownKey(ConfMonitoringServerMonitoringPath, "/liveness")
```

pkg/ffapi/apiserver_config.go
```diff
@@ -22,8 +22,9 @@ import (
 )
 
 var (
-	ConfMonitoringServerEnabled     = "enabled"
-	ConfMonitoringServerMetricsPath = "metricsPath"
+	ConfMonitoringServerEnabled        = "enabled"
+	ConfMonitoringServerMetricsPath    = "metricsPath"
+	ConfMonitoringServerMonitoringPath = "monitoringPath"
```
Chengxuan, line 27:
```suggestion
	ConfMonitoringServerLivenessPath = "livenessPath"
```

pkg/ffapi/apiserver.go
```diff
@@ -63,6 +63,7 @@ type apiServer[T any] struct {
 	handleYAML                bool
 	monitoringEnabled         bool
 	metricsPath               string
+	monitoringPath            string
```
Chengxuan, line 66:
```suggestion
	livenessPath            string
```

pkg/ffapi/apiserver.go
```diff
@@ -102,6 +103,7 @@ func NewAPIServer[T any](ctx context.Context, options APIServerOptions[T]) APISe
 		requestMaxTimeout:         options.APIConfig.GetDuration(ConfAPIRequestMaxTimeout),
 		monitoringEnabled:         options.MonitoringConfig.GetBool(ConfMonitoringServerEnabled),
 		metricsPath:               options.MonitoringConfig.GetString(ConfMonitoringServerMetricsPath),
+		monitoringPath:            options.MonitoringConfig.GetString(ConfMonitoringServerMonitoringPath),
```
Chengxuan, line 106:
```suggestion
		livenessPath:            options.MonitoringConfig.GetString(ConfMonitoringServerLivenessPath),
```

pkg/ffapi/apiserver.go
```diff
@@ -303,9 +310,10 @@ func (as *apiServer[T]) createMonitoringMuxRouter(ctx context.Context) *mux.Rout
 		panic(err)
 	}
 	r.Path(as.metricsPath).Handler(h)
+	r.HandleFunc(as.monitoringPath, hf.APIWrapper(as.emptyJSONHandler))
```
Chengxuan, line 313:
```suggestion
	r.HandleFunc(as.livenessPath, hf.APIWrapper(as.emptyJSONHandler))
```

Chengxuan:
@Brindrajsinh-Chauhan I like the idea, I suggest that we name the route liveness path. You might need to clean up the commit to be signed as well.

pkg/ffapi/apiserver_config.go
Brindrajsinh-Chauhan: Microservices like servicemanager uses it as both liveness and readiness probes. Wanted to keep is somewhat generic since using `/liveness` for readiness probe seems a bit odd

Chengxuan: I suggested the renaming based on my own experience with k8s apps, in general, a simple empty response could be sufficient for liveness checks, but often insufficient for readiness checks.

pkg/ffapi/apiserver.go
onelapahead: https://stackoverflow.com/a/43381061

Screenshot 2025-02-20 at 09 18 36

Some quick searching, and remembering from other OSS Go projects like NATS, that `/healthz`, `/livez`, and `/readyz` tend to be the convention which is apparently a Google best practice called z-pages.

I think `/livez` as the default (and maybe `/metricz` too ?) would match this convention better. I think its OK that its non-generic, its very clear then what this endpoint does: all it does is report up if the server is running, which is great as a basically health probe for Docker, K8s, Nomad, whatever.

Lib users can then implement their own routes/z-pages to represent various things about a service like its readiness to receive traffic and other internals.

pkg/ffapi/apiserver_config.go
```diff
@@ -49,4 +50,5 @@ func InitAPIServerConfig(apiConfig, monitoringConfig, corsConfig config.Section)
 	httpserver.InitHTTPConfig(monitoringConfig, 6000)
 	monitoringConfig.AddKnownKey(ConfMonitoringServerEnabled, true)
 	monitoringConfig.AddKnownKey(ConfMonitoringServerMetricsPath, "/metrics")
+	monitoringConfig.AddKnownKey(ConfMonitoringServerLivenessPath, "/healthz")
```
onelapahead, line 53: okay changed my mind: `/livez` I think to match what we've named this route and to be more modern. The beauty of it is that our users can set `livenessPath` if they want to be called something else like `/healthz` 🚀

onelapahead (approved):
sorry my spelling nitz, this is looking good to me and should reduce the boiler plate a lot

Chengxuan (approved):

EnriqueL8 (commented):
Looks good - just needs DCO checks to pass first

pkg/ffapi/apiserver.go
```diff
@@ -303,9 +310,10 @@ func (as *apiServer[T]) createMonitoringMuxRouter(ctx context.Context) *mux.Rout
 		panic(err)
 	}
 	r.Path(as.metricsPath).Handler(h)
+	r.HandleFunc(as.livenessPath, hf.APIWrapper(as.emptyJSONHandler))
 
 	for _, route := range as.MonitoringRoutes {
-		r.HandleFunc(route.Path, as.routeHandler(hf, route)).Methods(route.Method)
+		r.HandleFunc(fmt.Sprintf("/%s", route.Path), as.routeHandler(hf, route)).Methods(route.Method)
```
EnriqueL8, line 316: confused why there is a slash here? I see the default is `/livez`

EnriqueL8 (approved):
Thanks for making that change - good to go!

PR #172, opened 2025-03-25
[ffresty] Avoiding Leakable Rate Limiter Map on Resty Client Creation

onelapahead:
We found in a (bad) service that was concurrently/continously creating `resty.Client`'s via `ffresty` that:
1. concurrently calling `New` could lead to panics/crashes due to `fatal error: concurrent map writes`
2. the map of rate limiters per client would slowly leak memory over time since theres no way to clean up the rate limiter map entry on garbage collection of the client

Instead, we can use a piece of referenced memory for each client to keep track of it's respective rate limiter without needing a (concurrent friendly) map to track it, and Go's garbage collection will it ensures its properly deleted over time - if users are silly enough to create and discard `resty.Client`'s frequently - which is not recommended, but not necessarily unsupported.

### DIFF
--- a/pkg/ffresty/ffresty.go
+++ b/pkg/ffresty/ffresty.go
@@ -61,7 +61,6 @@ type Config struct {
 }
 
 var (
-	rateLimiterMap map[*resty.Client]*rate.Limiter
 	metricsManager metric.MetricsManager
 	onErrorHooks   []resty.ErrorHook
 	onSuccessHooks []resty.SuccessHook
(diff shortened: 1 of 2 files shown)


Chengxuan (approved):
nice, the initial intention was just to make the rate limiter per client scoped, so using a map to track all the clients is unnecessary compared to the approach here.

EnriqueL8 (approved):
Awesome, makes sense to me! Once the client is deleted then the GC will clean up that captured variable `rateLimiter` 👍🏼

PR #173, opened 2025-04-02
Allow use of the column "sequence" from consumers of dbsql

peterbroadhurst:
I have a use of this library where I need the column `sequence` to be an actual column, not the `SERIAL` tracker (which we generally support renaming to something else).

I can't change the default here, as that might break code depending on this slightly weird edge code.


.golangci.yml
```diff
@@ -9,11 +9,13 @@ linters-settings:
...
       - regexpMust
   goheader:
+    only-new-issues: true
     values:
       regexp:
         COMPANY: .*
+        YEAR_LAX: '202\d'
     template: |-
-      Copyright © {{ YEAR }} {{ COMPANY }}
+      Copyright © {{ YEAR_LAX }} {{ COMPANY }}
```
EnriqueL8, line 18: thanks for this

EnriqueL8 (commented):
Looks good - needs a small unit test

.golangci.yml
peterbroadhurst: See you've got dup changes up-stream - pulled that in

pkg/dbsql/crud_test.go
```diff
@@ -189,7 +189,8 @@ func newCRUDCollection(db *Database, ns string) *TestCRUD {
 			NilValue:     func() *TestCRUDable { return nil },
 			NewInstance:  func() *TestCRUDable { return &TestCRUDable{} },
 			ScopedFilter: func() squirrel.Eq { return sq.Eq{"ns": ns} },
-			EventHandler: nil, // set below
+			EventHandler: nil,  // set below
+			IDField:      "id", // same as default
```
peterbroadhurst, line 193: This was the unit test I included @EnriqueL8 - could you elaborate on the additional test you'd like?

awrichar (approved):

PR #174, opened 2025-04-07
feat: bump to 1.23 and minor fixes

SamMayWork:


Makefile
```diff
@@ -21,7 +21,7 @@ lint: ${LINT}
 ${MOCKERY}:
 		$(VGO) install github.com/vektra/mockery/v2@latest
 ${LINT}:
-		$(VGO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.55.2
+		$(VGO) install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.8
```
SamMayWork, line 24: Just to call this one out, I don't have any of the context for why this was pinned but it was causing build failures on `1.23` so I've bumped to the latest under the major version and it seems to be working fine.

EnriqueL8: thanks

go.mod
```diff
@@ -1,14 +1,14 @@
...
+go 1.23
 
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.2
 	github.com/Masterminds/squirrel v1.5.4
 	github.com/aidarkhanov/nanoid v1.0.8
 	github.com/docker/go-units v0.5.0
 	github.com/fsnotify/fsnotify v1.7.0
-	github.com/getkin/kin-openapi v0.122.0
+	github.com/getkin/kin-openapi v0.131.0
```
EnriqueL8, line 11: What changes did this introduce to generate swagger? Breaking changes apparently based on your below changes, will this affect end users?

EnriqueL8 (commented):
Thanks - looks good, one comment on the openapi bump

go.mod
SamMayWork: https://github.com/getkin/kin-openapi/pull/964

As fart as I can see the breaking changes

SamMayWork: Build seemed to work fine

EnriqueL8: Thanks for checking!

EnriqueL8 (approved):
Looks good!

pkg/ffapi/openapi3.go
```diff
@@ -188,21 +188,20 @@ func (sg *SwaggerGen) ffTagHandler(ctx context.Context, route *Route, name strin
...
 		schema.Format = "date-time"
 	case "Bytes32":
-		schema.Type = typeString
+		schema.Type = &openapi3.Types{openapi3.TypeString}
 		schema.Format = "byte"
 	case "FFBigInt":
-		schema.Type = typeString
+		schema.Type = &openapi3.Types{openapi3.TypeString}
 	case "JSONAny":
-		schema.Type = ""
```
EnriqueL8, line 205: Missed that this broke the jsonany swagger that is why I was seeing swagger changes in FireFly Core recently

PR #175, opened 2025-04-16
[ffapi] [ffresty] Allow for Configuring Request Header for Distributed Trace Logging

onelapahead:
In some cases, we might be integrating with existing distributed tracing systems. While `PassthroughHeaders` helps with that, it doesn't allow for per-req headers to be added/generated like a trace ID.

Instead - we can have the servers generate req IDs for any header and their clients will use that req ID in their context's for downstream requests.

### DIFF
--- a/pkg/ffapi/handler.go
+++ b/pkg/ffapi/handler.go
@@ -39,7 +39,24 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-const FFRequestIDHeader = "X-FireFly-Request-ID"
+const DefaultRequestIDHeader = "X-FireFly-Request-ID"
+
+var (
+	requestIDHeader string = DefaultRequestIDHeader
+)
+
+// RequestIDHeader returns the header name used to pass the request ID for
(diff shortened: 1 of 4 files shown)


pkg/ffapi/handler.go
```diff
@@ -39,7 +39,24 @@ import (
 	"github.com/sirupsen/logrus"
 )
 
-const FFRequestIDHeader = "X-FireFly-Request-ID"
```
onelapahead, line 42: Could avoid the breaking change and keep it the same name if we prefer @EnriqueL8

EnriqueL8: Uhm so consumers could be importing this field, I think we could release this as part of 1.5

EnriqueL8 (commented):
Looks good

EnriqueL8:
Quick search suggests it's just being used in TS and not in any Golang FireFly packages 
https://github.com/search?q=org%3Ahyperledger+FFRequestIDHeader&type=code

PR #180, opened 2025-05-05
Add support for wrapping an existing WS connection

peterbroadhurst:
The `wsclient` package is very helpful for wrapping a raw websocket connection, with heartbeating, and throttling.

I would like to use it for consistent handling like this of both outbound established connections, and inbound established connections (not using `wsserver` which is too opinionated for me).

This PR allows me to `Wrap()` an existing WebSocket per the demonstration unit test and treat it equivalently to a client connection.

> Also fixed an issue where no backoff factor was being set on the `retry.Retry` for connect retry, meaning the library would fall back to 1 (so no backoff)


onelapahead (approved):

EnriqueL8 (approved):
Looks good!

PR #183, opened 2025-05-22
Add bg WS connect to common utility - superseding initialConnectAttempts

peterbroadhurst:
The journey (multiple hops) that lead to https://github.com/hyperledger/firefly/pull/1315 told us that `initialConnectAttempts` was a bit of a flawed coupling strategy between asynchronously starting microservices.

We abandoned it, but left the feature in the common library without a clue that it was flawed.

So this PR tries to help other users like me from repeating the mistake :)

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -89,15 +89,10 @@ type sectionParent interface {
 	AddChild(key string, defValue ...interface{})
 }
 
-type SectionArrayCommon interface {
-	KeySet
-	SubSection(name string) Section
-}
-
 // Section represents a section of the global configuration, at a nested point in the config hierarchy.
 // Note that all keys are added to a GLOBAL map, so this cannot be used for per-instance customization.
 type Section interface {
-	SectionArrayCommon
+	KeySet
 	SetDefault(key string, defValue interface{})
 	SubSection(name string) Section
 	SubArray(name string) ArraySection
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -89,10 +89,15 @@ type sectionParent interface {
 	AddChild(key string, defValue ...interface{})
 }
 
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -89,10 +89,15 @@ type sectionParent interface {
...
+	SubSection(name string) Section
+}
+
 // Section represents a section of the global configuration, at a nested point in the config hierarchy.
 // Note that all keys are added to a GLOBAL map, so this cannot be used for per-instance customization.
 type Section interface {
-	KeySet
+	SectionArrayCommon
 	SetDefault(key string, defValue interface{})
 	SubSection(name string) Section
```
EnriqueL8, line 102: This is now superseded by SectionArrayCommon?

pkg/config/config.go
```diff
@@ -89,10 +89,15 @@ type sectionParent interface {
 	AddChild(key string, defValue ...interface{})
 }
 
+type SectionArrayCommon interface {
```
EnriqueL8, line 92: Does it need retrieval functions?

EnriqueL8: I know the experience for Array config hasn't been great and it seem like you are trying to improve this ?

EnriqueL8 (commented):
Looks good - but don't understand the config change, feels like a missing piece there on how that interface is used

EnriqueL8:
Merged a fix from @hosie and test coverage as well

pkg/config/config.go
peterbroadhurst: Sorry, this snuck in from another bit of work.

This definitely needs sorting, but wasn't for this PR will remove

EnriqueL8 (approved):
Looks good!

PR #185, opened 2025-05-28
Add ability to dynamically add filter JSON fields (and test coverage)

peterbroadhurst:
When processing a `JSONQuery` filter it is helpful to be able to dynamically build up a list of fields, for cases where there might be join tables to something very dynamic like a set of `tags`/`labels` that require an individual join for each filterable tag.

- Added `FieldResolver` plug extension following the existing plugin pattern for `ValueResolver`
- Allowed a `Clone()` of a field set, to a new map that lets to add fields
   - As per the included test you need to update this dynamically so the `Finalize()` will work
- `endsWith` had been omitted form JSON Query - so added that

## Additional things found

- We couldn't support array typed input on the `JSONHandler` due to a quirk where the type was ending up as `[]interface{}` rather than the type returned from `JSONInputValue`. I've proposed a new `JSONInputDecoder` for that, which lets you work around this very simply:
    - https://github.com/hyperledger/firefly-common/blob/fcaa664b46cd6d77936a1179e29524545431878e/pkg/ffapi/apiserver_test.go#L125-L129
- There was some invalid code in the filter logic around BigInt processing that I fixed writing missing tests - it was discarding numbers that are perfectly acceptable for big integer filters. Unlikely to have been used, but I think this was a copy/paste from the other code to make a compiler check on integer limits go away.
- Found some untested code around versions and OpenAPI constructs and validated it executed as written
- I was wondering about the `200` rather than `204` return code that was specified in the liveness check, which returns a `Content-Type: application/json` without any body. But I just commented that rather than changing it. 
    - Not sure if @onelapahead has a thought there... I was worried about changing it (doesn't block merging)

### DIFF
--- a/pkg/ffapi/apiserver.go
+++ b/pkg/ffapi/apiserver.go
@@ -386,9 +386,8 @@ func (as *apiServer[T]) notFoundHandler(res http.ResponseWriter, req *http.Reque
 	return 404, i18n.NewError(req.Context(), i18n.Msg404NotFound)
 }
 
-func (as *apiServer[T]) emptyJSONHandler(res http.ResponseWriter, _ *http.Request) (status int, err error) {
-	res.Header().Add("Content-Type", "application/json")
-	return 200, nil
+func (as *apiServer[T]) noContentResponder(res http.ResponseWriter, _ *http.Request) {
+	res.WriteHeader(http.StatusNoContent)
 }
 
 func (as *apiServer[T]) createMonitoringMuxRouter(ctx context.Context) (*mux.Router, error) {
@@ -400,7 +399,7 @@ func (as *apiServer[T]) createMonitoringMuxRouter(ctx context.Context) (*mux.Rou
 		panic(err)
 	}
 	r.Path(as.metricsPath).Handler(h)
-	r.HandleFunc(as.livenessPath, hf.APIWrapper(as.emptyJSONHandler))
+	r.HandleFunc(as.livenessPath, as.noContentResponder)
 
 	for _, route := range as.MonitoringRoutes {
 		path := route.Path
--- a/pkg/ffapi/apiserver.go
+++ b/pkg/ffapi/apiserver.go
@@ -202,7 +202,7 @@ func (as *apiServer[T]) Serve(ctx context.Context) (err error) {
 		if err != nil {
 			return err
 		}
-		as.monitoringPublicURL = buildPublicURL(as.MonitoringConfig, apiHTTPServer.Addr())
+		as.monitoringPublicURL = buildPublicURL(as.MonitoringConfig, monitoringHTTPServer.Addr())
 		go monitoringHTTPServer.ServeHTTP(ctx)
 	}
 
@@ -217,6 +217,10 @@ func (as *apiServer[T]) Started() <-chan struct{} {
 
 func (as *apiServer[T]) APIPublicURL() string {
 	return as.apiPublicURL
+}
+
+func (as *apiServer[T]) MonitoringPublicURL() string {
+	return as.monitoringPublicURL
 }
 
 func (as *apiServer[T]) waitForServerStop(httpErrChan, monitoringErrChan chan error) error {
--- a/pkg/ffapi/apiserver_test.go
+++ b/pkg/ffapi/apiserver_test.go
@@ -18,6 +18,7 @@ package ffapi
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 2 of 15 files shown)


EnriqueL8 (commented):
Looks good to me - I don't see the commented code you mentioned on the 200 /204

pkg/ffapi/apiserver_test.go
```diff
@@ -176,6 +200,50 @@ func TestAPIServerInvokeAPIRouteStream(t *testing.T) {
...
+
+func TestAPIServerInvokeAPIRouteLiveness(t *testing.T) {
+	_, as, done := newTestAPIServer(t, true)
+	defer done()
+
+	<-as.Started()
+
+	res, err := resty.New().R().Get(fmt.Sprintf("%s/livez", as.MonitoringPublicURL()))
+	assert.NoError(t, err)
+	assert.Equal(t, 200, res.StatusCode()) // note this should really be 204, but not changing as would change behavior
```
peterbroadhurst, line 235: This is the thing I was commenting on @onelapahead @EnriqueL8 - I _didn't_ change the behavior, but I did fill in the missing test and it felt like my test found a bug

onelapahead: These checks were designed for Kubernetes: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-http-request

> Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure.

So changing this status code to the more proper 204 would not be breaking change in behavior from that perspective.

EnriqueL8: @peterbroadhurst @onelapahead Is one of you changing it as part of this PR?

pkg/wsclient/wsclient.go
```diff
@@ -375,14 +375,18 @@ func (w *wsClient) connect(initial bool) error {
...
+				b, readErr := io.ReadAll(res.Body)
 				res.Body.Close()
+				if readErr == nil && len(b) > 0 {
+					// The info we need is what the server returned and the status
+					errMsg = string(b)
+				}
 				status = res.StatusCode
 			}
-			l.Warnf("WS %s connect attempt %d failed [%d]: %s", w.url, attempt, status, string(b))
+			l.Warnf("WS %s connect attempt %d failed [%d]: %s", w.url, attempt, status, errMsg)
```
peterbroadhurst, line 389: Because of the use of `Custom Log` this was failing to actually log the underlying error at all in the case the error wasn't data returned from the remote server (but something like a connection failure instead)

pkg/wsclient/wsclient.go
```diff
@@ -307,6 +307,10 @@ func (w *wsClient) Send(ctx context.Context, message []byte) error {
 		return nil
 	case <-ctx.Done():
 		return i18n.NewError(ctx, i18n.MsgWSSendTimedOut)
+	case <-w.sendDone:
```
peterbroadhurst, line 310: This was an edge case deadlock:

```
goroutine 12 [select]:
github.com/hyperledger/firefly-common/pkg/wsclient.(*wsClient).Send(0x40004d2100, {0xb86230, 0x40004be0a0}, {0x4000412770?, 0xffffffffffffffff?, 0x0?})
        /go/pkg/mod/github.com/hyperledger/firefly-common@v1.5.6-0.20250629160958-9871d4332701/pkg/wsclient/wsclient.go:305 +0xd4
----- CODE THAT DOES A SEND IN AN afterConnect() CALLBACK ----
github.com/hyperledger/firefly-common/pkg/wsclient.(*wsClient).receiveReconnectLoop(0x40004d2100)
        /go/pkg/mod/github.com/hyperledger/firefly-common@v1.5.6-0.20250629160958-9871d4332701/pkg/wsclient/wsclient.go:542 +0x194
created by github.com/hyperledger/firefly-common/pkg/wsclient.(*wsClient).initialConnect in goroutine 10
        /go/pkg/mod/github.com/hyperledger/firefly-common@v1.5.6-0.20250629160958-9871d4332701/pkg/wsclient/wsclient.go:254 +0x68
FAIL    github.com/kaleido-io/workflow-engine/pkg/enginesdk     10.024s
FAIL
```

onelapahead (approved):

EnriqueL8 (approved):
Looks good - just added explanation of one piece as it wasn't obvious

pkg/ffapi/apiserver.go
```diff
@@ -382,9 +386,8 @@ func (as *apiServer[T]) notFoundHandler(res http.ResponseWriter, req *http.Reque
 	return 404, i18n.NewError(req.Context(), i18n.Msg404NotFound)
 }
 
-func (as *apiServer[T]) emptyJSONHandler(res http.ResponseWriter, _ *http.Request) (status int, err error) {
-	res.Header().Add("Content-Type", "application/json")
-	return 200, nil
```
EnriqueL8, line 387: This is technically a breaking change... so might need a minor version bump

PR #189, opened 2025-09-15
Add ability to explicitly set a field to NULL

peterbroadhurst:
Without this PR, if you tried to do `.Set("fieldName", nil)` you would get:
```
FF00143: Unable to parse value for filter 'fieldName'
```


EnriqueL8 (approved):
Looks good - thanks @peterbroadhurst !

PR #190, opened 2025-09-17
Update Swagger UI and allow JSON queries to specify `fields`

peterbroadhurst:
- Updates Swagger UI lib to to 5.29.0 - felt like time to get an upgrade proposed
- Adds `fields` which were missing in the JSON query payload, to specify what fields should be returned
    - Already supported in the underlying filter capabilities


EnriqueL8 (approved):
Looks good!

PR #192, opened 2025-10-05
JSON query builder

peterbroadhurst:
- In PR chain with #190 

Adds a useful Go structure to allow a much more compact way for Go code to build JSON query structures:

https://github.com/hyperledger/firefly-common/blob/153db91fe6c539cff9bee637f42bba673869f294/pkg/ffapi/restfilter_json_builder_test.go#L90-L104

> Note this ports back a piece of work that was done in a separate LFDT OSS project (Paladin) that uses the same JSON format for querying.


EnriqueL8 (approved):
Looks really good, this is gonna make it a lot easier - copyright date needs updating

PR #194, opened 2025-10-18
Fix trace level logging

peterbroadhurst:
Trace level logging current looks like this for binary data, which isn't very helpful for problem debug:
```
[2025-10-18T11:59:45.411-04:00] TRACE ==> (body) [123 34 110 97 109 101 34 58 34 115 58 ... 
```

This PR improves this, although it doesn't go as far as providing full trace for streamed `io.Reader` payloads.


EnriqueL8 (approved):
Thanks @peterbroadhurst !

PR #195, opened 2025-10-23
Add ability to exclude a specific route from the OpenAPI and expose SwaggerHTML

EnriqueL8:
New `ExcludeFromOpenAPI` to exclude a route from showing up in the OpenAPI and expose the SwaggerHTML instead of only allowing to access it through a `HandlerFunction`


peterbroadhurst (approved):
👍

PR #198, opened 2025-11-03
do not mutate scoped filter

hosie:
`idFilter` was modifying the scoped filter on a collection with an `id` filter condition and any attempts read from that collection afterwards, would always be scoped to a single record.

This PR changes that so we now only read from the scoped filter and do not modify it.


peterbroadhurst (approved):
👍

PR #199, opened 2025-11-16
Add option for explicit nested AND sub-queries

peterbroadhurst:
I hit a scenario that I couldn't describe with our JSONQuery syntax.

See in this test, we have a top-level AND where one of the sub-queries is an OR. This couldn't be described before.

https://github.com/hyperledger/firefly-common/blob/8f24fbfd4a04c9f2ec80b80ac1dd1b8e2908611f/pkg/ffapi/restfilter_json_test.go#L145-L176

### DIFF
--- a/pkg/i18n/en_base_field_descriptions.go
+++ b/pkg/i18n/en_base_field_descriptions.go
@@ -43,8 +43,8 @@ var (
 	FilterJSONSkip               = ffm("FilterJSON.skip", "Number of results to skip before returning entries, for skip+limit based pagination")
 	FilterJSONSort               = ffm("FilterJSON.sort", "Array of fields to sort by. A '-' prefix on a field requests that field is sorted in descending order")
 	FilterJSONCount              = ffm("FilterJSON.count", "If true, the total number of entries that could be returned from the database will be calculated and returned as a 'total' (has a performance cost)")
-	FilterJSONOr                 = ffm("FilterJSON.or", "Array of sub-queries where any sub-query can match to return results (OR combined). Note that within each sub-query all filters must match (AND combined)")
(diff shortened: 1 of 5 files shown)


pkg/i18n/en_base_field_descriptions.go
```diff
@@ -43,7 +43,8 @@ var (
 	FilterJSONSkip               = ffm("FilterJSON.skip", "Number of results to skip before returning entries, for skip+limit based pagination")
 	FilterJSONSort               = ffm("FilterJSON.sort", "Array of fields to sort by. A '-' prefix on a field requests that field is sorted in descending order")
 	FilterJSONCount              = ffm("FilterJSON.count", "If true, the total number of entries that could be returned from the database will be calculated and returned as a 'total' (has a performance cost)")
-	FilterJSONOr                 = ffm("FilterJSON.or", "Array of sub-queries where any sub-query can match to return results (OR combined). Note that within each sub-query all filters must match (AND combined)")
+	FilterJSONOr                 = ffm("FilterJSON.or", "Array of sub-queries")  // Note due to complex issue in swagger generator AND/OR need to be identical
+	FilterJSONAnd                = ffm("FilterJSON.and", "Array of sub-queries") // ^^^ the `description` field is getting pushed to the sub-schema definition, and is non-deterministic
```
peterbroadhurst: What happens in the swagger generator is this:
```yaml
        or:
          items:
            $ref: '#/components/schemas/FilterJSON'
          type: array
```

Then depending on non-deterministic behavior, the `FilterJSON` schema gets a `description` from _ether` the `and` or `or` definition.

That makes swagger generation non-deterministic.

IMHO this is a bug, and the swagger should look like this - but that would be a hard contribution to build+validate to the openapi3gen package:

```yaml
        or:
          description: "the description of the reference to the object"
          items:
            $ref: '#/components/schemas/FilterJSON'
          type: array
```

EnriqueL8 (approved):
Looks great @peterbroadhurst  - left a few suggestions of typos and question on raising issue.

PR #200, opened 2025-11-19
[log] logr.Logger and logr.LogSink Support for controller-runtime Compatibility

onelapahead:
Kubernetes controllers love `logr` as an interface for logging.

FireFly loves `logrus` (log levels are so much better than `-vvvvv`).

🍅 🥔 

In our Kubernetes operators - we'd prefer to provide the controller-runtime / K8s libraries with an ff-common like logger, so that we can then use `log.L(ctx)` semantics w/ proper log levels to create a smoother logging experience for our developers vs our users.

These wrapper implementations would help achieve that.


EnriqueL8 (commented):
Thanks @onelapahead  - this is awesome 🙇🏼 

Left a few comments/things I don't understand

PR #205, opened 2026-01-08
[dbsql] [ffapi] DB Stats and Metrics

onelapahead:
~Go metrics can be useful when needing to understand heap usage and how much GC is consuming resources.~ UPDATE: go metrics were already enabled, we just instead enabled build info metrics and have ffapi tests to ensure the metrics endpoint works as expected.

DB metrics are in two forms:
1. stats collector wraps `Stats()` from `*sql.DB` with basic gauges around conn settings and pool usage
2. instrumentation of all operations and transactions

Similar to `ffresty`, there's then a global func for clients to enable DB metrics, ensuring all initialized databases in a process are metered by the collector and instrumentation.


pkg/dbsql/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+// EnableDBMetrics registers a prometheus collector for the given database connection.
+func EnableDBMetrics(_ context.Context, metricsRegistry metric.MetricsRegistry, db *sql.DB, options ...MetricsOption) {
+	opts := &metricsOptions{}
+	opts.apply(options...)
+
+	if opts.dbName == "" {
+		opts.dbName = "default"
+	}
+
+	metricsRegistry.MustRegisterCollector(collectors.NewDBStatsCollector(db, opts.dbName))
```
EnriqueL8: Info on stats https://pkg.go.dev/database/sql#DBStats

pkg/ffapi/apiserver.go
```diff
@@ -306,6 +310,11 @@ func (as *apiServer[T]) createMuxRouter(ctx context.Context) (*mux.Router, error
 	}
 
 	if as.monitoringEnabled {
+		if as.goProcessMetricsEnabled {
+			as.MetricsRegistry.MustRegisterCollector(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
+			as.MetricsRegistry.MustRegisterCollector(collectors.NewGoCollector())
+		}
+		as.MetricsRegistry.MustRegisterCollector(collectors.NewBuildInfoCollector())
```
EnriqueL8: Important comment on this collector:
```
The labels will only have meaningful values if the binary is built with Go module support and from source code retrieved from the source repository (rather than the local file system).
```

pkg/ffapi/apiserver.go
```diff
@@ -65,6 +66,7 @@ type apiServer[T any] struct {
 	alwaysPaginate            bool
 	handleYAML                bool
 	monitoringEnabled         bool
+	goProcessMetricsEnabled   bool
```
EnriqueL8: This is whole pkg is already Golang 

```suggestion
	processMetricsEnabled   bool
```

EnriqueL8 (commented):
Thanks @onelapahead - this is great! 

A few comments

pkg/ffapi/apiserver.go
onelapahead: Yeah debatable if we want this..

pkg/ffapi/apiserver.go
onelapahead: Removed as these were actually already enabled.

pkg/dbsql/metrics.go
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// EnableDBMetrics registers a prometheus collector for the given database connection.
+func EnableDBMetrics(_ context.Context, metricsRegistry metric.MetricsRegistry, db *sql.DB, options ...MetricsOption) {
+	opts := &metricsOptions{}
+	opts.apply(options...)
+
+	if opts.dbName == "" {
+		opts.dbName = "default"
+	}
```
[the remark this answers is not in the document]
onelapahead: Instead - I parse the DB name out of the URL we provide on initialization, and allow for providers to implement their own parsers if needed.

pkg/dbsql/metrics.go
```diff
@@ -0,0 +1,119 @@
...
+// EnableDBMetrics registers prometheus metrics for database operations.
+// Must be called before any Database.Init to ensure metrics are wired up.
+func EnableDBMetrics(ctx context.Context, _metricsRegistry metric.MetricsRegistry) (err error) {
+	if metricsRegistry != nil {
+		log.L(ctx).Warn("DB metrics are already enabled, skipping")
+		return nil
+	}
+	metricsRegistry = _metricsRegistry
+
+	metricsManager, err = metricsRegistry.NewMetricsManagerForSubsystem(ctx, subsystem)
```
EnriqueL8, line 57: should the subsystem name be configurable?

EnriqueL8 (commented):
Looks good - just one comment here

PR #207, opened 2026-02-09
add support for oneof tags

hosie:
- Added support for oneof tags that can reference other known schemas by name
- Added static function to Register a schema from a golang struct.  Can be called statically (during init). Most likely this call would be co-located to the definition of the type struct.
- Added Unregister function, primarily to help with unit testing but a consuming microservice may also decide to dynamically enable / disable certain structs as potential targets for oneof references.
- and oneOf resolution support in Swagger generation, including an option to panic on unresolved oneOf references.  Default is not to panic which makes it safe to call for dynamic generated swagger and the caller can detect unresolved references using a validator after generation. By leaving this option as false, caller also has the option to inject the referenced Schema's after generation (test TestOneOfWithPostInjectionOfSchema illustrates this approach)

### DIFF
--- a/pkg/ffapi/openapi3.go
+++ b/pkg/ffapi/openapi3.go
@@ -26,6 +26,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/getkin/kin-openapi/openapi3"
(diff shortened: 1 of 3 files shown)


peterbroadhurst (approved):
Lots of thinking on this one, but just one `fmt.Errorf()` requested for you to address.

I am worried `fftypes.JSONAny` will drive complexity in the calling code, and I see we already have some functionality to help with that:
https://github.com/hyperledger/firefly-common/blob/c1fccfd5aa31157e1cabc049daa190334579e0f1/pkg/fftypes/jsonany.go#L76

... but I'm interested in what approach you actually recommend there. I see there's a sample struct...

https://github.com/kaleido-io/firefly-common/blob/410b76f0bfe0c0d0487242a3d56a7121c04d92f6/pkg/ffapi/openapi3_test.go#L76-L78

But here there's no discriminator field for you to use to tell which of the two structs you should use when unmarshaling.... maybe it makes sense to round out the sample?

hosie:
FWIW, I ended up abandoning this approach because it resulted in more complexity than I liked on the golang struct and esp. with the custom JSONMarshal / Unmarshal.

PR #208, opened 2026-02-17
Add support for `ffschemaext`

chrisbygrave:
This PR adds support for a gostruct tag `ffschemaext` that allows a comma-separated list of key/value pairs to be added as extensions to the schema generated into the OpenAPI document. Keys must be of the format `x-<name>`.

For example, the following structure:
```go
type TestExtensions struct {
	MyString string `ffstruct:"mystruct" json:"mystring" ffschemaext:"x-key=value"`
}
```

Would generate a schema:
```
mystring:
  description: mystruct.mystring
  type: string
  x-key: value
```

Query string syntax (and escaping rules) can be used to add multiple extension to a single field:

```go
type TestExtensions struct {
	MyString string `ffstruct:"mystruct" json:"mystring" ffschemaext:"x-key=value&x-key2=value2"`
}
```


pkg/ffapi/openapi3.go
```diff
@@ -74,7 +74,10 @@ type BaseURLVariable struct {
 	Description string
 }
 
-var customRegexRemoval = regexp.MustCompile(`{(\w+)\:[^}]+}`)
+var (
+	customRegexRemoval = regexp.MustCompile(`{(\w+)\:[^}]+}`)
+	ffExtensionKeyRegexp = regexp.MustCompile(`^x-.+$`)
```
EnriqueL8: Are there any characters not support in the OpenAPI schema for keys? 
```suggestion
    // Check ffExtension key starts with "x-"
	ffExtensionKeyRegexp = regexp.MustCompile(`^x-.+$`)
```

EnriqueL8 (commented):
Looks good - a question and needs gofmt formatting

EnriqueL8 (approved):
Looks good to me @chrisbygrave  - thanks!

peterbroadhurst (commented):
Couple of thoughts here @chrisbygrave

chrisbygrave:
Worth noting that if the same property is specified multiple times, the last value will win - I think that's cleaner behaviour than introducing some arbitrary join character in an attempt to consolidate multiple values down to one.

peterbroadhurst (approved):
Thanks @chrisbygrave for going round the loop on the spelling and escaping semantics

PR #209, opened 2026-03-05
Allow ff tag annotations to be used with schema generation, outside SwggerGen

peterbroadhurst:
There are cases where being able to use `openapi3gen.NewSchemaRefForValue()` directly is very useful, but still exploiting all the value from `ffstruct` etc. tags that have been documented on `struct`s via the FF-common framework.

This PR also allows `ExportComponentOpts` from `openapi3gen` to be optionally specified in the `SwaggerGenOptions`


pkg/ffapi/openapi3.go
```diff
@@ -66,8 +66,10 @@ type SwaggerGenOptions struct {
 	RouteCustomizations func(ctx context.Context, sg *SwaggerGen, route *Route, op *openapi3.Operation)
 
 	// OpenAPI 3.0.x specific options
-	Tags         openapi3.Tags
-	ExternalDocs *openapi3.ExternalDocs
+	Tags                       openapi3.Tags
+	ExternalDocs               *openapi3.ExternalDocs
+	ExportComponentOpts        *openapi3gen.ExportComponentSchemasOptions
+	AdditionalSchemaCustomizer func(name string, t reflect.Type, tag reflect.StructTag, schema *openapi3.Schema) error
```
EnriqueL8, line 72:
```suggestion
	AdditionalSchemaModifier func(name string, t reflect.Type, tag reflect.StructTag, schema *openapi3.Schema) error
```

pkg/ffapi/openapi3.go
```diff
@@ -351,11 +373,7 @@ func (sg *SwaggerGen) addURLEncodedFormInput(ctx context.Context, op *openapi3.O
 // CheckObjectDocumented lets unit tests on individual structures validate that all the ffstruct tags are set,
 // without having to build their own swagger.
 func CheckObjectDocumented(example interface{}) {
-	(&SwaggerGen{
-		options: &SwaggerGenOptions{
-			PanicOnMissingDescription: true,
-		},
-	}).Generate(context.Background(), []*Route{{
```
EnriqueL8, line 358: Feels like the expanded version was nicer to read

pkg/ffapi/openapi3.go
EnriqueL8, line 72: I don't see a test for any of these changes to add extract functionallity, I also wonder if these options should be nested into OpenAPI3GenOptions... This struct is becoming quite messy with a lot of options

EnriqueL8 (commented):
Looks good  - a few comments

pkg/ffapi/openapi3.go
EnriqueL8: This can be added in the future

EnriqueL8 (approved):

PR #211, opened 2026-03-17
Context DB txn timeout, and interface to avoid connection leak during migration

peterbroadhurst:
- In chain with #209

This addresses two issues with the DB driver:

1. When we `Begin` we're not supplying the context to allow the DB to automatically terminate the transaction when the context expires
2. When using PostgreSQL with auto-migrate, we're leaking a single connection for the life of the process, because the `WithInstance` function would close the _whole DB_ if we used `driver.Close()`
    - Solution is to use `WithConnection()` in the migration driver for PostgreSQL
    - Have to make this optional, as SQLite does not have a `WithConnection()`


pkg/dbsql/database.go
```diff
@@ -184,7 +194,7 @@ func (s *Database) BeginOrUseTx(ctx context.Context) (ctx1 context.Context, tx *
 	ctx1 = log.WithLogger(ctx, l)
 	before := time.Now()
 	l.Tracef("SQL-> begin")
-	sqlTX, err := s.db.Begin()
+	sqlTX, err := s.db.BeginTx(ctx /* transaction should auto-rollback on context cancel */, nil /* default options */)
```
onelapahead, line 197: 👍🏻 - this is what Opus was whispering in my ear about when analyzing this previously. Thanks makes perfect sense

pkg/dbsql/database.go
```diff
@@ -139,8 +140,17 @@ func (s *Database) RunAsGroup(ctx context.Context, fn func(ctx context.Context)
...
+	var driver database.Driver
+	providerClosable, isClosable := provider.(ProviderCloseableMigrationDriver)
+	if isClosable {
+		driver, err = providerClosable.GetMigrationDriverClosable(s.db)
+		defer func() {
+			_ = driver.Close()
+		}()
+	} else {
+		driver, err = provider.GetMigrationDriver(s.db)
+	}
```
onelapahead, line 153: Interesting - think this explains behavior I saw in some of our tests where when closing the migrations, it closed the whole driver, and caused other obscure failures in the tests 😵

onelapahead (approved):

EnriqueL8 (approved):
Awesome, makes sense to me!

PR #216, opened 2026-04-28
expose option for openapi version

hosie:
When using certain JSON schema constructs ( such as specifying type as an array), the generated API document will fail validation because those syntax were introduces in JSON Schema draft 2019-09.  OpenAPI 3.0.x aligns with earlier JSON Schema drafts where type MUST be a single string.

Open API 3.1.1 aligned with JSON Schema 2020-12 where type may be a string or an array of unique strings  .

So rather than hard coding the openapi version, this PR allows it to be configurable, but continues to default to `3.0.2`

### DIFF
--- a/pkg/ffapi/openapi3.go
+++ b/pkg/ffapi/openapi3.go
@@ -44,6 +44,8 @@ type SwaggerGenOptions struct {
 	Version     string
 	Description string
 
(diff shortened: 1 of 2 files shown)


EnriqueL8 (approved):
Thanks @hosie

PR #217, opened 2026-06-05
fixing enum compare to be case insensitive for FFEnumParseString

Chengxuan:
Among all functions apart from `Value()` , the usage of enum values follows a case-insensitive pattern and treats lower case as canonical. 

This PR updates the FFEnumParseString to also be case-insensitive when looking up a string against registered values.

### DIFF
--- a/pkg/fftypes/enum.go
+++ b/pkg/fftypes/enum.go
@@ -27,15 +27,23 @@ import (
 
 type FFEnum string
 
-var enumValues = map[string][]interface{}{}
+var enumValues = map[string][]string{}
 
 func FFEnumValue(t string, val string) FFEnum {
+	t = strings.ToLower(t)
 	enumValues[t] = append(enumValues[t], val)
 	return FFEnum(val)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/fftypes/enum.go
```diff
@@ -74,8 +74,12 @@ func FFEnumParseString(ctx context.Context, t, val string) (FFEnum, error) {
 		return "", i18n.NewError(ctx, i18n.MsgInvalidEnum, t)
 	}
 	for _, possible := range e {
-		if possible == strings.ToLower(val) {
-			return FFEnum(strings.ToLower(val)), nil
+		possibleStr, ok := possible.(string)
```
peterbroadhurst: Just `(string)possible` simple type casting should be possible here for a known type, rather than interface

Chengxuan: I see. conversion to interface{} is only needed for FFEnumValues as that function is used for openapi. 

Updated the code to make that clear.

peterbroadhurst (approved):
👍

PR #218, opened 2026-06-09
feat(fftls,ffresty,ffdns,ffnet) Networking Metrics and IP/Server Configs

onelapahead:
## fftls

Like `ffresty` and other pacakges - a static registration of a metrics registry makes a shared metrics manager subsystem and a single metric with `issuer`, `subject`, and `type` labels. Where the `float64` value of the gauge is the Unix timestamp of the expiry date (common pattern in Prom metrics like in FF DX metrics and Besu/Erigon timeSinceLastBlock).

## ffdns

A new config and resolver builder for customizing DNS timeouts and (name)servers for lookups. With metrics if enabled (though I don't think most users want DNS metrics always on as its expensive).

## ffnet

A custom `Dialer` with a configurable CIDR denylist for IP ranges to reject. Defaults to empty. Users can configure it to default to specific IANA IP ranges they may need to deny based on their use case (internal service-to-service client can be more permissive than a user-configurable webhooks client needing full SSRF protection).

## ffresty

Builds on `ffdns` and `ffnet` to embed their configs respectively like it does `fftls`.

Note that previously we never prescribed netgo - letting the compile time of the runtime determine what DNS implementation is used. When `ffdns` (timeout, servers, or metrics) is used, netgo must be used and is therefore specified.


calbritt:
Approved

pkg/ffresty/config.go
```diff
@@ -81,6 +81,10 @@ const (
 	// HTTPMaxIdleConnsPerHost the max number of idle connections per host
 	HTTPMaxIdleConnsPerHost = "maxIdleConnsPerHost"
 
+	// HTTPDNSServers an optional list of DNS server addresses (host or host:port, port defaults to 53) to use for
```
EnriqueL8: Are there any default DNS server addresses built into GO Dns resolver already?

pkg/ffresty/ffresty.go
```diff
@@ -376,6 +385,55 @@ func NewWithConfig(ctx context.Context, ffrestyConfig Config) (client *resty.Cli
...
+	}
+	return NewDNSResolver(ffrestyConfig.DNSServers, time.Duration(ffrestyConfig.HTTPConnectionTimeout))
+}
+
+// NewDNSResolver builds a pure-Go *net.Resolver that dials the given DNS servers
+// (each host or host:port, port defaulting to 53) in order, failing over to the
+// next on error. Returns nil when no servers are given (use the system resolver).
+// Exported so non-ffresty dialers — e.g. a WebSocket dialer — can honour the same
+// dnsServers config as the HTTP client.
+func NewDNSResolver(dnsServers []string, dialTimeout time.Duration) *net.Resolver {
```
EnriqueL8: this might fee like a separate file?

pkg/ffresty/ffresty.go
```diff
@@ -376,6 +385,55 @@ func NewWithConfig(ctx context.Context, ffrestyConfig Config) (client *resty.Cli
...
+// dnsResolver derives the resolver to attach to the dialer based on config precedence:
+//   - an explicitly provided programmatic Resolver always wins
+//   - otherwise, if DNSServers are configured, a pure-Go resolver that dials those servers
+//     (in order, failing over to the next on error) is built
+//   - otherwise nil, leaving Go's default system resolver selection in place
+func dnsResolver(ffrestyConfig *Config) *net.Resolver {
+	if ffrestyConfig.Resolver != nil {
+		return ffrestyConfig.Resolver
+	}
+	return NewDNSResolver(ffrestyConfig.DNSServers, time.Duration(ffrestyConfig.HTTPConnectionTimeout))
```
EnriqueL8: would we want the connection timeout to be a individually configurable?

pkg/ffresty/ffresty.go
```diff
@@ -376,6 +385,55 @@ func NewWithConfig(ctx context.Context, ffrestyConfig Config) (client *resty.Cli
...
+}
+
+// NewDNSResolver builds a pure-Go *net.Resolver that dials the given DNS servers
+// (each host or host:port, port defaulting to 53) in order, failing over to the
+// next on error. Returns nil when no servers are given (use the system resolver).
+// Exported so non-ffresty dialers — e.g. a WebSocket dialer — can honour the same
```
EnriqueL8: Yeah this exported reason makes it a different file or even pkg

EnriqueL8 (commented):
Thanks @onelapahead ! A few comments

pkg/ffresty/config.go
onelapahead: sorta - they just use the `/etc/resolv.conf` of the underlying OS (assuming Unix) and the equivalent for Windows by default.

So netgo is its own DNS resolver, but in terms of what nameservers to use it still gets that from the host, so that it operates w/ the proper control prescribed by the network admin.

Thats how Go microservices in K8s (using netgo 98% of the time in my personal experience) just magically know how to get the cluster CoreDNS IP address(es) for DNS resolution w/o any additional config.

pkg/ffresty/ffresty.go
onelapahead: eek I see - maybe `ffdns`

pkg/ffnet/config.go
```diff
@@ -0,0 +1,88 @@
...
+	"127.0.0.0/8",       // IPv4 loopback
+	"169.254.0.0/16",    // IPv4 link-local, incl. cloud metadata 169.254.169.254
+	"224.0.0.0/4",       // IPv4 multicast
+	"240.0.0.0/4",       // IPv4 reserved (incl. 255.255.255.255 broadcast)
+	"::1/128",           // IPv6 loopback
+	"::/128",            // IPv6 unspecified
+	"fe80::/10",         // IPv6 link-local
+	"fd00:ec2::254/128", // AWS IMDS IPv6 endpoint (cloud metadata)
+	"ff00::/8",          // IPv6 multicast
+}
```
onelapahead: Note - by default we do not include private IPs, bc for most use cases ffnet/ffresty is being used for private IP connectivity.

The burden is on the user to restrict it further to private IPs or whatever expected ranges, or relax it if loopback hosts are expected as a different example.

Disclaimer - we always recommend zero trust (mTLS) and/or high-trust (firewall) networking beneath these clients as the primary line of defense.

peterbroadhurst: I think where we provide a pre-baked config that can be enabled, it should be an enum to switch off _all private IP ranges_, including link-local.

I don't think it should be a default for all uses of `ffresty`.

onelapahead: Agreed - so we do not default to denying any, but there are now a bunch of IANA lists of IPs that `ffnet` users can import to build their own denylist's easily.

pkg/ffnet/config.go
```diff
@@ -0,0 +1,88 @@
...
+	AdditionalDeniedCIDRs = "additionalDeniedCIDRs"
+)
+
+// DefaultDeniedCIDRs are blocked by default to mitigate SSRF: loopback, link-local (including
+// the cloud metadata endpoint 169.254.169.254 and the AWS IMDS IPv6 endpoint), unspecified /
+// "this host", and multicast / reserved / broadcast ranges. Private RFC1918 / IPv6 ULA ranges
+// are intentionally NOT included — these dialers are commonly used for legitimate internal
+// service-to-service calls, so blocking private space is deferred to network firewalls /
+// zero-trust rather than baked in (callers wanting that can use additionalDeniedCIDRs).
+var DefaultDeniedCIDRs = []string{
```
EnriqueL8: Can these not be pulled in from `net/netip` or `netip.Addr` I see functions for IsLoopback, IsPrivate etc..  ?

pkg/ffnet/ffnet.go
```diff
@@ -0,0 +1,93 @@
+// Copyright © 2025 Kaleido, Inc.
```
EnriqueL8: nit: 2026

pkg/ffnet/ffnet_test.go
```diff
@@ -0,0 +1,204 @@
+// Copyright © 2025 Kaleido, Inc.
```
EnriqueL8: ditto

pkg/ffnet/ffnet_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+var utConf = config.RootSection("net_unit_tests")
+
+func resetConf() {
+	config.RootConfigReset()
+	InitConfig(utConf)
+}
+
+func TestNewDialControlDefaults(t *testing.T) {
+	// nil CIDRDenylist => secure defaults
```
EnriqueL8: are you moving towards an `enabled: true` flag in the config for the defaults?

pkg/ffdns/ffdns_test.go
```diff
@@ -0,0 +1,200 @@
+// Copyright © 2025 Kaleido, Inc.
```
EnriqueL8: ditto

pkg/ffnet/ffnet_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"224.0.0.0/4",          // IPv4 multicast
+	"240.0.0.0/4",          // IPv4 reserved (incl. 255.255.255.255 broadcast)
+	"fc00::/7",             // IPv6 ULA
+	"fe00::/8",             // IPv6 private RFC4193
+	"ff00::/8",             // IPv6 reserved
+	"::ffff:127.0.0.1/128", // IPv4-mapped IPv6 loopback
+	"::1/128",              // IPv6 loopback
+	"::/0",                 // IPv6 unspecified
+	"::/128",               // IPv6 unspecified
+}
```
peterbroadhurst, line 49: Really like this doc via UT approach 👍

peterbroadhurst (approved):
Looks great - thanks for working through the iterations.

pkg/ffnet/config.go
peterbroadhurst: Resolved with a doc-via-UT approach 👍

pkg/ffnet/config.go
```diff
@@ -0,0 +1,88 @@
+// Copyright © 2025 Kaleido, Inc.
```
EnriqueL8:
```suggestion
// Copyright © 2026 Kaleido, Inc.
```

pkg/ffnet/ffnet.go
EnriqueL8:
```suggestion
// Copyright © 2026 Kaleido, Inc.
```

pkg/ffnet/ffnet_test.go
EnriqueL8:
```suggestion
// Copyright © 2026 Kaleido, Inc.
```

EnriqueL8 (commented):
Looks good, some copyright updates

EnriqueL8 (approved):
Thanks you @onelapahead !

PR #219, opened 2026-06-23
Rename go.mod from hyperledger/firefly-common to hyperledger-firefly/common

dan-r:
Updating Go.mod after transferring the repo to new org


EnriqueL8 (approved):
Thanks @dan-r  - this is the beginning of moving pkgs to new org

PR #222, opened 2026-06-26
chore: upgrade to Go 1.26.4

dan-r:
Updates common to Go 1.26.4.

- `go.mod` go directive `1.23.0` → `1.26.4`
- CI `go-version` `1.23` → `1.26.4`
- Fix two non-constant format strings in tests that Go 1.26's `vet` now rejects (`pkg/fswatcher/fswatcher_test.go`, `pkg/i18n/errors_test.go`)

golangci-lint (v1.64.8) and all tests pass on Go 1.26.4. Note: bumping the module `go` directive requires downstream consumers of `hyperledger-firefly/common` to build with Go 1.26.4+.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/hyperledger-firefly/common
 
-go 1.23.0
+go 1.26.4
 
 require (
 	github.com/DATA-DOG/go-sqlmock v1.5.2
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v4
         with:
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/hyperledger-firefly/common
 
-go 1.23.0
+go 1.26.0
```
EnriqueL8: latest is 1.26.4 - can you upgrade to that please

EnriqueL8 (approved):
We need to put in the release notes that this completely breaking and that if anyone upgrade to the new version of common they will break

go.mod
EnriqueL8: arrgh I did this for CVE and was unideal..

PR #227, opened 2026-06-30
fix: invalid enum on container schemas in OpenAPI generation

EnriqueL8:
The ffenum tag handler applied schema.Enum unconditionally. For container fields ([]FFEnum, map[string]FFEnum) the openapi3gen library invokes the schema customizer with the same field tag for both the container schema and its element schema, so the enum was stamped on the "array"/"object" container node as well as the string leaf. An enum on a type:array or type:object node is invalid OpenAPI and is rejected by strict SDK generators (kin-openapi's own Validate() does not catch it).


peterbroadhurst (approved):
Thanks @EnriqueL8 - this looks very clear to me

onelapahead (approved):
Can we add an openapi linter utility ?

EnriqueL8:
We can add an openapi linter in a following PR, looked into different package vaccuum could be one https://github.com/daveshanley/vacuum

PR #238, opened 2026-07-24
fftls: added ClientTLSConfigMergingSystemCAs for HTTP clients that ne…

IvanNik:
### Summary
- Add TLS config option `includeSystemCAs` (default `false`).
- When a custom CA (`caFile` / `ca`) is set **and** `includeSystemCAs: true`, merge that CA into the system cert pool instead of replacing `RootCAs`.
- Existing replace-only behavior is unchanged when the flag is unset/false.

### Config
| Key | Default | Effect when `caFile`/`ca` set |
|-----|---------|-------------------------------|
| `includeSystemCAs` | `false` | `false`: custom CA only (existing). `true`: system pool + custom CA. |

Example:
```yaml
tls:
  enabled: true
  caFile: /path/to/private-ca.pem
  certFile: /path/to/client.pem
  keyFile: /path/to/client-key.pem
  includeSystemCAs: true
```

### Security
- Opt-in only; default preserves CA pinning.
- Does not skip hostname verification.
- Prefer for outbound clients that may hit public HTTPS redirects; avoid widening inbound `ClientCAs` trust unless intentional.

### DIFF
--- a/pkg/fftls/config.go
+++ b/pkg/fftls/config.go
@@ -44,7 +44,13 @@ const (
 	// HTTPConfTLSRequiredDNAttributes provides a set of regular expressions, to match against the DN of the client. Requires HTTPConfTLSClientAuth
 	HTTPConfTLSRequiredDNAttributes = "requiredDNAttributes"
 
-	defaultHTTPTLSEnabled = false
+	// HTTPConfTLSIncludeSystemCAs when true and a custom CA (caFile/ca) is configured, merges that CA
+	// into the system certificate pool instead of replacing RootCAs. Useful for outbound HTTP clients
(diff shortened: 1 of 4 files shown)


onelapahead (approved):
Looks great! Thanks Ivan

EnriqueL8 (commented):
Thanks @IvanNik - build failing 
```
Error: pkg/fftls/fftls.go:208:81: newRootCAPool - result 1 (error) is always nil (unparam)
func newRootCAPool(ctx context.Context, includeSystemCAs bool) (*x509.CertPool, error) {
```

EnriqueL8 (approved):
Thanks @IvanNik  for the changes

